PR #11, opened 2021-05-19
feat: benchmark specification I3

### LINKED ISSUE #5: I3: Specification - Percent of genome-wides sites identified
Create specification for I3.

**Working definition:** Compare output of tool's RNA-seq-based site identification to established atlas(es) of polyadenylation sites (PolyASite, PolyA_DB, GENCODE, RefSeq).

**Notes:** This would be a measure of what percent of sites identified are near (within a fixed window) of a database's sites.

daneckaw:
Adds I3 specification (Compare output of tool's RNA-seq-based site identification to established atlas(es) of polyadenylation sites). 
closes #5 

Adds files:
- benchmark specification: `challenges\identification\I3_benchmark\I3_specification.md`
- example output files: 
  - `challenges\identification\I3_benchmark\example_files\output1.json`
  - `challenges\identification\I3_benchmark\example_files\output2.json`


challenges/identification/I3_benchmark/I3_specification.md
```diff
@@ -0,0 +1,113 @@
...
+Based on the input data the following metrics are computed:
+
+1. Precision = (TP/(TP+FP))
+
+TP - true positives - PAS identified by the tool and present within X nucleotides of PAS identified in the database  
+FP - false positives  - PAS identified by the tool and not present within X nucleotides of PAS identified in the database
+
+The metrics should be calculated for the following thresholds of distance between PAS identified by the tool and PAS present in databases:
+
+- 50 nt
```
mfansler, line 22: Do we also want this expanded to a sequence from 0 to 200 nt using 10 nt steps? Or maybe 20 nt steps?

challenges/identification/I3_benchmark/I3_specification.md
```diff
@@ -0,0 +1,113 @@
...
+
+* **Challenge:** Identification
+* **Identifier:** I3
+
+## Inputs
+
+| # | Format | Link | Example data |
+  | --- | --- | --- | --- |
+  | 1 | BED | [Specification][spec-bed] | [Link][in1] |
+  | 2 | BED | [Specification][spec-bed-polyAsite] | [Link][in2] |
```
mfansler, line 37: Would it be useful to have these pre-conformed? That is, should we have adapters that conform polyASite and PolyA_DB to standard BED6 format? It doesn't seem like we'll use any of the other data.

challenges/identification/I3_benchmark/I3_specification.md
```diff
@@ -0,0 +1,113 @@
...
+  | 5 | Precision_100nt_PolyAdb | N/A | Output 2 | Read file, parse JSON and extract attribute `Precision_100nt` | `vector` | N/A |
+  | 6 | Precision_200nt_PolyAdb | N/A | Output 2 | Read file, parse JSON and extract attribute `Precision_200nt` | `vector` | N/A |
+  
+### Additional info metrics
+  
+  N/A
+
+[//]: # (References)
+  
+  [in1]: ./example_files/input1.bed
```
mfansler, line 103: Can we add versions of the input files? `input1.bed` can be same as before for I2. If the inputs stay as polyASite and PolyA_DB formats, we can just dump a `head -n10` or so for them.

mfansler (commented):
I'm okay approving as is, as long as we make an issue for adding the input files.

challenges/identification/I3_benchmark/I3_specification.md
daneckaw: For the other ones we've used 0-100nt with 10nt step - we can do the same here or is there a reason to use 0-200nt range (and should we change it to 200 for other benchmarks too)?

challenges/identification/I3_benchmark/I3_specification.md
daneckaw: If we want to add pre-conformed example inputs, then we'd also have to modify actual database files before for the workflow to run, right? I think it would be better to just leave it for the evaluation teams to process the database files inside the workflow. **Or** we could add a separate issue for database file processing and then this benchmark would be database-agnostic.

challenges/identification/I3_benchmark/I3_specification.md
mfansler: I'm fine either way. We could change I2 to 200 nt just to be consistent. Although, I kind of expect 100 nt on the ground truth might be sufficiently informative since it's coming from the same biological sample, whereas here we are comparing against other experimental results.

challenges/identification/I3_benchmark/I3_specification.md
mfansler: @daneckaw I sent a PR to your branch with these.

daneckaw:
Ready to merge if approved

challenges/identification/I3_benchmark/I3_specification.md
mfansler: I was thinking the latter, but we could get to it when it comes up.

mfansler (approved):

mfansler:
@uniqueg this one's good now 👍

challenges/identification/I3_benchmark/I3_specification.md
daneckaw: I'm thinking now, should we have distances in a separate input file so the code is a bit more flexible?

ninsch3000:
@mfansler @daneckaw Is it possible that you're adding 3 files twice, once into a directory `challenges`, which doesn't exist on the main anymore because it has been replaced by `summary_workflows` (where you're also adding the same files, plus input files)? Or is it just me not being able to interpret github correctly (again)?

mfansler (approved):

PR #16, opened 2021-05-21
feat: benchmark specification Q2

### LINKED ISSUE #8: Q2: Specification - Correlation to 3'-End Seq
Create specification for Q2.

**Working definition:** Compute correlation coefficient between tool's quantification obtained RNA-seq and the quantification from the corresponding 3'-end sequencing dataset (for each pair of test datasets).

daneckaw:
Adds Q2 specification (Correlation to 3'end seq quantification). 
closes #8 

Adds files:
- benchmark specification: `challenges\quantification\Q2_benchmark\Q2_specification.md`
- example output files: 
  - `challenges\quantification\Q2_benchmark\example_files\output1.json`


mzavolan:
Will have to define how sites from the RNA-seq and 3' end seq are mapped to each other.

The distance for mapping sites between data sets should be a parameter.

dominikburri:
Just to be clear, the BED has 6 columns, and I assume column 5 is for TPM? 
At least this would make the BED usable in IGV browser, e.g.

Issue #64, opened 2021-05-28
Matching up predicted with ground truth sites


mzavolan:
Both for identification and for quantification we need to identify which ground truth site corresponds to each predicted site, if a 'site' is not a single base. We assume that ground truth sites, whether single position or interval, are correct and non-overlapping. The script should take as input bed files for both predicted and ground truth data, as well as a distance parameter, specifying by how many nucleotides should the predicted sites be extended to the  left and right. Predicted sites that overlap can be merged and their expression summed up. For each predicted site/cluster the group truth sites that overlap with it can then be identified. If a predicted site/cluster overlaps with more than one ground truth cluster,  it can be assigned to each of the overlapping ground truth sites with a fractional weight. This could be calculated as the number of overlapping nucleotides divided by the length of the predicted site/cluster. The expression can assigned proportionally to these weights. 

To have the full information for  computing various metrics the output should contain:
* 1. ID, start and end of predicted site
* 2. ID, start and end of overlapping ground truth site
* 3. weight with which the predicted site was assigned to the ground truth site
* 4. expression of the predicted site assigned to the ground truth site
* 5. expression level of ground truth site

mfansler:
Related: #8 and #9

mrgazzara:
After our discussion @ninsch3000 found [bedtools window](https://bedtools.readthedocs.io/en/latest/content/tools/window.html) as a promising approach to accomplish the distance based overlaps between BED files.

PR #66, opened 2021-05-31
feat: scipts to match predicted sites to ground truth

lschaerfen:
Fixes issues #64 and should remove blocker from #8 

* `summary_workflows/quantification/match_with_gt.py` uses `bedtools window` to match predicted and ground truth sites. Expression values are weighted in case multiple predicted sites overlap one ground truth, or _vice versa_.
* `summary_workflows/quantification/corr_with_gt.py` takes the output (predicted sites matched with ground truth) and calculates the correlation coefficient of the corresponding expression levels.
* `summary_workflows/quantification/corr_with_gt.py` takes the output (predicted sites matched with ground truth) and calculates the correlation coefficient of the corresponding expression levels.
* Updates `summary_workflows/quantification/README.md` to explain usage of scripts.


ninsch3000:
Thanks a lot @lschaerfen , looks like great work to me, and will surely be useful in more than one benchmark! Unfortunately I'm a much less versed programmer than you, so wouldn't want to give advice. Maybe @mzavolan and @mrgazzara  could review?

summary_workflows/quantification/match_with_gt.py
```diff
@@ -0,0 +1,187 @@
...
+## sum expression levels for predicted sites that overlap with each other given the window size
+## assuming the prediction bed file has exactly 6 columns in the correct order, no header!
+
+# find sites that overlap within the predict set itself when the window is extended
+out = bedtools_window(f_PD, f_PD, window)
+
+# for logging
+len_dup = len(out)
+
+not_unq_mask = out.duplicated([0, 1, 2, 5], keep=False)
```
mzavolan, line 63: Having numbered columns makes it difficult to read the code. Would have been better if the columns had meaningful names.

summary_workflows/quantification/match_with_gt.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # collapse non-unique sites, sum up expression, add back to bed file
+    unq = np.unique([(i, j, k, l) for i, j, k, l in zip(not_unq[0], not_unq[1], not_unq[2], not_unq[5])], axis=0)
+
+    for i in unq:
+        non_unq_context = not_unq[(not_unq[0]==i[0]) & (not_unq[1]==int(i[1])) & (not_unq[2]==int(i[2])) & (not_unq[5]==i[3])]
+
+        merged_start = min(list(non_unq_context[7]))
+        merged_end = max(list(non_unq_context[8]))
+        merged_exp = sum(non_unq_context[10])
```
mzavolan, line 83: Here it seems that the expression of all ground truth sites that overlap with a predicted site is summed up. Is this intended?

summary_workflows/quantification/README.md
```diff
@@ -0,0 +1,48 @@
...
+The script `match_with_gt.py` uses bedtools window to assign ground truth PAS to the predictions. Each prediction is extended by `n` base pairs, depending on the parameter chosen for `window`.
+
+The script `corr_with_gt.py` calculates the correlation coefficient between prediction and matched ground truth quantification values. The input must be a BED file containing the columns below. Usage: `python3 corr_with_gt.py <prediction_merged.bed>`
+
+Processing steps:
+- run bedtools window for prediction file with itself
+- if necessary, merge prediction sites that fall into the window, expression is summed
+- run bedtools window for merged predictions with ground truth file
+- find multiple ground truth sites overlapping one predicted site
+	- weigths are added for the predicted site based on distance to the ground truth site
```
mzavolan, line 12: Should be more specific: the weight is the fraction of the predicted site overlapping with the ground truth site (according to the implementation)

mzavolan: However, if a predicted site overlaps with multiple ground truth sites, shouldn't rather the predicted site be split?

mzavolan (changes requested):

summary_workflows/quantification/match_with_gt.py
daneckaw: I'm not sure if I fully understand the code, but I think this part relates to summing up expression of all predicted PAS that are within the same window (`bedtools window` is run on predicted PAS both as input A and input B).

lschaerfen: Yes exactly! This was @mzavolan's suggestion in issue #64. Hope I understood that correctly.

daneckaw:
Is there a reason why the predicted PAS are merged before mapping to ground truth? I can see a couple of problems, especially for bigger windows: distant predicted PAS could be merged into one and then mapped to two different ground truth PAS, or predicted PAS that is more than one window away from GT could be first merged with a site located closer to GT. Or does the code work differently and I'm misunderstanding something?

summary_workflows/quantification/match_with_gt.py
daneckaw: I see, it makes sense. But I'm wondering if it's necessary in this case when we have both a window for merging and for mapping to GT? Maybe it should be one or the other? Or at least maybe the windows should have different values, i.e. one small, one bigger?

lschaerfen:
You're not misunderstanding, this is how the code works. The first step is to merge sites that fall within the window parameter. I can add a parameter to turn off merging, so we can figure out later if we want that or not?

summary_workflows/quantification/match_with_gt.py
```diff
@@ -0,0 +1,187 @@
...
+
+        out = out.append({0: i[0], 1: merged_start, 2: merged_end, 3: list(non_unq_context[3])[0], 4: merged_exp, 5: i[3]}, ignore_index=True)
+
+    # de-duplicate, sort
+    out.drop_duplicates([0, 1, 2, 5], inplace=True)
```
daneckaw, line 88: Duplicates are removed but I'm not sure if it solves the problem if multiple different merged sites are created. For example when we have sites at locations: 4, 8, 12, 16 with window of size 6, we will end up with several overlapping sites, right?

summary_workflows/quantification/match_with_gt.py
mzavolan: > Yes exactly! This was @mzavolan's suggestion in issue #64. Hope I understood that correctly.
> 
> It seems like it's the ground truth sites, but it's actually the predicted sites because bedtools window is called with the same input file twice. I'll make that more readable.

Ah, yes, it makes sense, missed that trying to resolve indices. The comments are there, but could be clearer.

mzavolan: > I see, it makes sense. But I'm wondering if it's necessary in this case when we have both a window for merging and for mapping to GT? Maybe it should be one or the other? Or at least maybe the windows should have different values, i.e. one small, one bigger?

Yes, this relates to my comment about splitting predicted sites that match multiple ground truth sites. In principle, we trust the ground truth sites to be precise.

lschaerfen: We can have a different window for merging. What would you suggest? 

@mzavolan would you suggest splitting one predicted site that matches multiple ground truth sites into equal parts?

summary_workflows/quantification/match_with_gt.py
lschaerfen: I'm not 100% sure I understand the question. Can you describe the overlap scenario you think can cause a problem? 

The de-duplication is because artificial duplicates are added by the for loop above. I found this was easier than checking every time in the loop if weighting on one PAS had already been performed.

mzavolan: bedtools window will only find extended sites from one file overlapping with files from the other file. But I don't think it constructs single-linkage clusters of overlapping sites. So yes, I think we would end up with multiple overlapping sites from the predictions file.

Going back to the original issue, to try to deal with some degree of noise in 3' end processing, I would suggest to not merge the predictions, but take them face-value as individual sites. One can extend them (but a small number of nucleotides) when searching for overlaps with the ground truth. Then, the following scenarios can occur:
1. extended predicted site does not overlap with any ground truth site - the ground truth site expression can be set to 0 (the rest of the fields can be set from the predicted site)
2. extended predicted site overlaps with one ground truth site - sites are matched with each other
3. extended ground truth site overlap with multiple ground truth sites - if included, the expression of this site should be split between the ground truth sites (to make it clear the name can acquire a suffix and the genomic interval can be split, but depending on the functions used, it may be OK to have duplicate name and coordinates). How to split? Simplest would be in proportion to the overlap of the genomic intervals (as currently done in the code).

daneckaw: In a scenario with PAS at: 10, 20, 30 and 10 nt window we'd end up with merged PAS like so: 10-20, 10-30 and 20-30. If each of individual sites had TPM=20, then cummulative TPM of merged sites would be: 40, 60, 40. Now assuming GT sites are at: 5 and 35, then mapping would be: 

- 10-20 site mapped to GT-5 with weight=1, 
- 10-30 site mapped to GT-5 with weight=0.5, 
- 10-30 site mapped to GT-35 with weight=0.5,
- 20-30 site mapped to GT-35 with weight=0.5

The cummulative TPM of predicted sites mapped to GT-5 would be then 40 * 1 + 60 * 0.5 = 70, same for GT-35. When in reality it should be closer to 30 for each (20 from the closest predicted site, and maybe 10 from the site between them).

I don't know how likely a similar scenario is, but I think in some edge cases, depending on tools output or window size, it can be problematic.

(If I got something really wrong here, I'm sorry for the essay)

lschaerfen:
I addressed your first two points @mzavolan in the latest commit. For now sites that fall under your 3rd point are discarded. I will look into your other points later, but at least it is usable now. In the test data not many sites fall into that category when using a reasonable window parameter.

summary_workflows/quantification/README.md
```diff
@@ -0,0 +1,48 @@
...
+	3. end prediction
+	4. name prediction
+	5. expression prediction
+	6. strand prediction
+	7. chromosome ground truth
+	8. start ground truth
+	9. end ground truth
+	10. name ground truth
+	11. expression ground truth (additional columns go after this one, such as ground truth gene_ID)
+	12. weight of prediction expression
```
dominikburri, line 29: To make it compatible with BED format, I would include "strand ground truth" in column 12. Then the output is BED of prediction, BED of ground truth and then additional columns.

summary_workflows/quantification/corr_with_gt.py
```diff
@@ -0,0 +1,40 @@
...
+
+fname = args.bed
+out = pd.read_csv(fname, delimiter='\t', header=None)
+
+# initialize vectors for ground truth sites and prediction
+vec_true = []
+vec_pred = []
+
+# multiple predicted sites for one ground truth?
+multiple_predicted_sites = out.duplicated([6, 7, 8, 11], keep=False)
```
dominikburri, line 20: I think it was mentioned already, but for better readability it would be nice to have named columns.

summary_workflows/quantification/corr_with_gt.py
```diff
@@ -0,0 +1,40 @@
...
+from scipy.stats import pearsonr
+import json
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument('bed', help='The BED file containing predictions merged with ground truth.')
+args = parser.parse_args()
+
+fname = args.bed
+out = pd.read_csv(fname, delimiter='\t', header=None)
```
dominikburri, line 13: Here you could read in the csv with specific column names. As this is known and fixed, I don't see a reason to hardcode this.

summary_workflows/quantification/match_with_gt.py
```diff
@@ -0,0 +1,133 @@
...
+import numpy as np
+from shutil import which
+import argparse
+
+parser = argparse.ArgumentParser()
+
+parser.add_argument('a', help='The BED file containing predictions. MUST be BED6 format.')
+parser.add_argument('b', help='The ground truth bed file. First 6 columns must be standard BED6, but can have additional columns appended.')
+parser.add_argument('window', help='Number of bases to append to each side of the predicted site.', type=int)
+#parser.add_argument('-o', help='output file directory') # not yet implemented!!
```
dominikburri, line 13: Instead of output directory, you could also give the option to save output to specified file name.

summary_workflows/quantification/match_with_gt.py
```diff
@@ -0,0 +1,133 @@
...
+args = parser.parse_args()
+
+
+f_GT = args.b
+f_PD = args.a
+window = args.window
+
+
+def bedtools_window(bed1, bed2, window, reverse=False):
```
dominikburri, line 22: Could you explain the parameter reverse? According to bedtools the parameter "-v" means "Only report those entries in A that have no overlaps with B.", is this correct?

summary_workflows/quantification/match_with_gt.py
```diff
@@ -0,0 +1,133 @@
...
+import pandas as pd
+import numpy as np
+from shutil import which
+import argparse
+
+parser = argparse.ArgumentParser()
+
+parser.add_argument('a', help='The BED file containing predictions. MUST be BED6 format.')
+parser.add_argument('b', help='The ground truth bed file. First 6 columns must be standard BED6, but can have additional columns appended.')
+parser.add_argument('window', help='Number of bases to append to each side of the predicted site.', type=int)
```
dominikburri, line 12: Might be easier to rename `window` to `window_size`.

summary_workflows/quantification/match_with_gt.py
```diff
@@ -0,0 +1,133 @@
...
+        
+        # weights = np.array(olps)/np.sum(olps)
+        # non_unq_context['weight'] = weights
+        
+        # # add back the removed sites including weights
+        # out = out.append(non_unq_context)
+
+# find sites with no overlap given the window
+out_rev = bedtools_window(f_PD, f_GT, window, reverse=True)
+out_rev.rename({0: 'chrom_p', 1: 'chromStart_p', 2: 'chromEnd_p', 3: 'name_p', 4: 'score_p', 5: 'strand_p', 6: 'chrom_g', 7: 'chromStart_g', 8: 'chromEnd_g', 9: 'name_g', 10: 'score_g', 11: 'strand_g'}, axis=1, inplace=True)
```
dominikburri, line 116: You could do the renaming in `bedtools_window` directly, then you don't have to write the same code twice.

dominikburri (commented):
Mostly housekeeping comments.

summary_workflows/quantification/match_with_gt.py
lschaerfen: Yes, this is to find the polyA sites that do not have an overlap in the ground truth set.

PR #87, opened 2021-06-02
APA-Scan & Roar worflow

EuancRNA:
…e to size and upload constraints

Fixes #{ISSUE_NUMBER}

## Checklist

- [x] I have tested my code


EuancRNA:
NOTE THAT THIS IS COMBINED WITH WHAT WAS MEANT TO BE A SEPARATE PULL REQUEST FOR ROAR AS WELL. APOLOGIES IM NOT USED TO WORKING WITH GIT AND GITHUB

EuancRNA:
APA-Scan should be fully working with all envs and configs etc, except for the fasta file (mm10.fa for the test data) which I remove. Will need to readjust elements of samples.csv and config etc to match the exact directories and files for the actual data to be processed, not the test.

Roar should work except I wasn't able to get the docker image working as Ican't get docker on this PC due to not having admin rights. There is a docker image for it publically available, which can be found here: http://quay.io/biocontainers/misopy:0.5.4--py27heb79e2c_4 . Other than that there's a conda yaml file Roar.yaml that has every package you need except for roar itself, which can be downloaded as a bioconductor package here: https://bioconductor.org/packages/release/bioc/html/roar.html

execution_workflows/APA-Scan/APA-Scan.py
```diff
@@ -0,0 +1,116 @@
+
```
uniqueg, line 1: This looks like the APA-Scan code itself, no? If so, I think this should be put inside a Docker container, copied straight from its original repo, no need to keep a copy here. Goes for all other APA-Scan scripts as well, e.g., `make_plots.py`, `methods.py` etc.

You can check out the Dockerfile in the mountainClimber PR to get a rough idea on how to do that (by cloning the Git repo and then copying the Python scripts): https://github.com/iRNA-COSI/APAeval/blob/ewf_mountainclimber/execution_workflows/mountainClimber/mountainClimber_nextflow_dsl2/docker/Dockerfile_method

execution_workflows/APA-Scan/README.md
```diff
@@ -0,0 +1,24 @@
+# APA-Scan
```
uniqueg, line 1: I think this should focus on how to run the execution workflow rather than the tool. In particular, it should include a description of the various config files, sample sheets etc

execution_workflows/APA-Scan/configuration.ini
```diff
@@ -0,0 +1,29 @@
+[INPUT_RNAseq]
+# Input folder names
+# All samples(names like sample1_1.bam, sample1_2.bam....) in group1 must be inside of one folder, put the diectory on input1 parameter
+# All samples(names like sample1_1.bam, sample2_2.bam....) in group2 must be inside of one folder, put the diectory on input2 parameter
+input1 = /home/bs14e3m/projects/other/APAeval/execution_workflows/APA-Scan/srsf3
```
uniqueg, line 5: Not good to put absolute file paths of your own system in a Git repo

execution_workflows/APA-Scan/config/samples.csv
```diff
@@ -0,0 +1,3 @@
+sample,sample_type,data_type
```
uniqueg, line 1: Not sure how, but I think run-/sample-specific configs should be organized in some way to make it easy to run specific sets of matching configs (with each run relating to a parameter code in the list we prepared). It's okay to keep this as a dummy example though I think

execution_workflows/APA-Scan/workflow/Snakefile
```diff
@@ -0,0 +1,357 @@
+"""
+Snakemake pipeline for APA-Scan.
+For help see: https://snakemake.readthedocs.io/en/stable/index.html.
+"""
+
+
+
+
+
+##### READ THIS ###############################################################
```
uniqueg, line 10: Okay, guess I should have read that first. Not sure how to deal with that though. Are you absolutely sure you didn't change the way that APA-Scan works? Because otherwise we'd be benchmarking _you_ rather than the published tool and that might put us in a line of fire. In any case, the better approach would be to fork their repository, add the required changes, then build a Docker image that contains your fork rather than the original. I would also open a PR to the original repo from that fork, letting the authors know about your changes, why you did them and what they would fix. This would at least give them the chance to respond to them, so that we can hopefully - eventually - run the workflow with a version that is sanctioned by the authors

execution_workflows/APA-Scan/workflow/Snakefile
```diff
@@ -0,0 +1,357 @@
+"""
```
uniqueg, line 1: Would be good to have the Snakemake experts review this, e.g., @mkatsanto, @dominikburri, @ninsch3000

execution_workflows/APA-Scan/workflow/envs/APA-Scan.Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Template Dockerfile with python and pandas.
```
uniqueg, line 1: See other comments on how to get the APA-Scan code inside the container as well, then update the Snakefile to use Docker rather than Conda for cloud execution

execution_workflows/APA-Scan/workflow/rules/[SUBWORKFLOW].smk
```diff
@@ -0,0 +1 @@
+# Place any additional SUBWORKFLOWS here.
```
Comment (uniqueg) [inline execution_workflows/APA-Scan/workflow/rules/[SUBWORKFLOW].smk:1]: To keep things clean, delete any template folders/files you don't need. Applies to more files further down as well

execution_workflows/Roar/.Rhistory
```diff
@@ -0,0 +1,2 @@
+library(biomaRt)
```
uniqueg, line 1: Like `.RData` this should not go inside the repo. In fact these may be good candidates to add to `.gitignore`

execution_workflows/Roar/dryrun.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+snakemake \
+    --rerun-incomplete \
+    --snakefile="workflow/Snakefile" \
+    --configfile="config/config.APA-Scan.yaml" \
```
uniqueg, line 5: Should probably be from Roar?

execution_workflows/Roar/README.md
```diff
@@ -0,0 +1,24 @@
+# APA-Scan
```
uniqueg, line 1: Most comments on APA-Scan apply to Roar as well. Perhaps it would indeed be good to prepare another branch for Roar as this PR is quite unwieldy and some copy/paste artefacts crept in, so it's a bit risky to merge these two features into one PR. Please ask on Slack if you need help to disentagle that

execution_workflows/Roar/rulegraph.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+snakemake \
+    --snakefile="workflow/Snakefile" \
+    --configfile "config/config.APA-Scan.yaml" \
```
uniqueg, line 4: Roar instead of APA-Scan

execution_workflows/Roar/run_local.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+snakemake \
+    --snakefile="workflow/Snakefile" \
+    --configfile="config/config.APA-Scan.yaml" \
```
uniqueg, line 4: Roar instead of APA-Scan

execution_workflows/Roar/workflow/Snakefile
```diff
@@ -0,0 +1,141 @@
+"""
```
uniqueg, line 1: Again, would be good to have @dominikburri, @ninsch3000 and/or @mkatsanto have a look at this

execution_workflows/Roar/workflow/envs/APA-Scan.Dockerfile
```diff
@@ -0,0 +1,25 @@
+# Template Dockerfile with python and pandas.
```
uniqueg, line 1: File name includes APA-Scan. Also, it doesn't look like anything Roar-specific is being done in this Dockerfile

execution_workflows/Roar/workflow/envs/APA-Scan.yaml
```diff
@@ -0,0 +1,64 @@
+name: APA-Scan
```
uniqueg, line 1: Should remove this file

execution_workflows/Roar/workflow/envs/roar.yaml
```diff
@@ -0,0 +1,245 @@
+name: roar
```
uniqueg, line 1: How do `roar.yaml` and `Roar.yaml` differ?

execution_workflows/Roar/workflow/scripts/Roar_polya.R
```diff
@@ -0,0 +1,109 @@
+
```
uniqueg, line 1: Where is this from? Did you write that? Would be good to comment

execution_workflows/Roar/workflow/scripts/Roar_polya.Rmd
```diff
@@ -0,0 +1,127 @@
+---
```
uniqueg, line 1: I think this can be removed

uniqueg (changes requested):
Thanks a lot Euan - quite some work! :))
Reviewed APA-Scan only for now. Had a few comments, mostly will involve removing files and something related to Docker, so should be not too hard. Once the changes are in, I think we can have a more in-depth review of the Snakefiles and scripts you wrote :)

I was also wondering about the origin of this file:
* `execution_workflows/APA-Scan/ncbiRefSeq.txt`: What is this required for? Is it part of our standardized inputs? If not, is it compatible with them?

execution_workflows/APA-Scan/workflow/Snakefile
```diff
@@ -0,0 +1,357 @@
...
+# THE LAYOUT OF THESE WILL DICTATE THE ORIENTATION OF LEVELS THAT ARE COMPARED
+# BY APA-SCAN. IE FOR THE TEST DATA, THIS MAKES SO SRSF3 is CONDITION1 AND
+# THE CONTROL IS CONDITION 2
+
+################################################################################
+
+
+
+# Parameters
+THREADS = 8
```
ninsch3000, line 130: same as below... better define in config.
Ideally the whole workflow could be run with different samples, annotations, outputnames, threads, ...., anything that comes to mind, WITHOUT changing the code anywhere.

execution_workflows/APA-Scan/workflow/Snakefile
```diff
@@ -0,0 +1,357 @@
...
+# attempt to make it a bit generalisable (touch wood). I'm hoping whatever levels
+# you want to compare, you could parse that in some way from the previous step?
+
+################################################################################
+
+
+
+rule generate_configurationfile:
+    """A rule that does some preprocessing.
+    This example copies input to output and writes stdout and stderr to log.
```
ninsch3000, line 202: change this rule description. Notes from the paragraph above could go here (in a shortened form maybe)

execution_workflows/Roar/README.md
ninsch3000: I agree, I'm afraid the two methods should really be on separate PRs

ninsch3000 (changes requested):
That looks like a lot of work @EuancRNA , thanks a lot!
I do agree with @uniqueg though, all the input and output and other unnecessary files should be removed from the repo, and the two methods should have separate branches/PRs. I also agree that a fork from the original repo should be created in case you need to change the code of the method itself.
Furthermore I suggest to parameterize some values, see my comments.
For more technical comments I suggest review from @dominikburri  and @mkatsanto !

execution_workflows/APA-Scan/configuration.ini
EuancRNA: So I should have put the explaination in the file itself, the configuration.ini is run-specific, and is created by editing the blank_configuration.ini file in a rule in the Snakefile. The snakemake script is written to replace a placeholder (for input1 its input_dir1, with the current working dir of the pipeline. Wasn't 100% sure if this would work on the cloud but can just relplace the os.getcwd() with whatever dir you want. APA-Scan.py also wouldnt seem to work without specifying an absolute path hence the need for it here.

execution_workflows/APA-Scan/workflow/Snakefile
EuancRNA: Yeah I've tried multiple times, the only think I could think is to use the exact directories they specify - but I'm not able to do hthat on my work PC because I cant make new dirs in root. Cheers for the advice.

PR #91, opened 2021-06-02
Add Q2 benchmark specification and sample files

### LINKED ISSUE #8: Q2: Specification - Correlation to 3'-End Seq

daneckaw:
Fixes #8 

Added Q2 benchmark specification and sample input and output files. Parts related to PAS mapping between RNAseq and orthogonal dataset copied from GH-66 README file.

## Checklist

- [ ] I have tested my code


summary_workflows/quantification/Q2_benchmark/Q2_benchmark_specification.md
```diff
@@ -0,0 +1,99 @@
...
+### Additional info inputs
+
+#### Format 1
+
+This BED file contains single-nucleotide positions of unique poly(A) sites as well as TPM values for each identified site quantified from RNAseq data by the benchmarked tool.  
+Fields:
+
+- **chrom** - the name of the chromosome
+- **chromStart** - the starting position of the feature in the chromosome
+- **chromEnd** - the ending position of the feature in the chromosome; as identified PAS are single-nucleotide, the ending position is the same as starting position
```
ninsch3000, line 48: Is this true for bed format? Isn't end exclusive?

lschaerfen: I believe a single nucleotide site would still have chromEND = chromStart +1.

mfansler: I think this coheres. From the spec:

> chromStart and chromEnd can be identical, creating a feature of length 0, commonly used for insertions. For example, use chromStart=0, chromEnd=0 to represent an insertion before the first nucleotide of a chromosome.

So, if we have `chromStart=10, chromEnd=10, strand='+'`, that would represent cleaving after position 9 and before position 10.

mfansler (approved):

summary_workflows/quantification/Q2_benchmark/Q2_benchmark_specification.md
ninsch3000: But a single nucleotide is not an insertion? I'm not convinced ;)

mfansler: The cleavage site is not at a nucleotide but between them, so I think the BED files are correct. Maybe the use of the word "single-nucleotide" is the issue? Perhaps more simply:

```
- **chromStart** - the position of cleavage in genomic coordinates
- **chromEnd** - the position of cleavage in genomic coordinates
```

where, as a zero-based coordinate system, interprets as:
 - `0` == 5' end of first base
 -  `1` == 3' end of first base == 5' end of second base
 -  ...

Does that make sense?

lschaerfen: Yes that does make sense! Are you sure that the tools also report it in this way? At least for the ground truth sites that I've been using for testing, the sites are always reported as a 1-nt interval.

ninsch3000: You're absolutely right @mfansler , makes perfect sense! But I also agree with Leo, there might be developers as ignorant as me who report a single nucleotide. But then again, some tools might report clusters (ranges), as we did for the ground truth. I think Matthew converted those to the representative site we have in the identifier. But what will we do if a tool only reports clusters? Will that be an issue? Do we need checks or transformations before the bedtools window script? Or will it not matter?

mfansler: @lschaerfen I wasn't aware the ground truth are 1-nt. But still I think what is here should stand as _the specification_ - tools that output something different should be conformed to it. 

@ninsch3000 For clusters, some of us had discussed using the center. I'm of the opinion that should be done upstream of this, so that this benchmark could validate that an input BED specifies cleavage sites with 0-nt, but it wouldn't be responsible for transforming it. I think the evaluation workflows would be responsible for deciding how to translate a particular tool's output to a cleavage site specification.

lschaerfen: The bedtools window tool should currently handle clusters, the given interval will be extended by the window parameter. This of course makes less precise tools perform better I guess. 

I have not tested it with 0-nt coordinates, might give an error. I'll go back and check.

daneckaw: Single-nucleotide sites were requested in the same form in execution workflows output specification. Do we need different wording and maybe some clarification for execution workflow team?

lschaerfen: Yeah maybe. I was assuming the 1-nt interval was default since I found that in the ground truth mock data. But maybe I used a too early version of those files...

mfansler: @daneckaw I think so. I added a note in this branch on using the 0-nt specification. If you think that helps clarify it, we could copy it to the output specification.

mfansler:
@ninsch3000 @mrgazzara @lschaerfen I propose we move this forward as is. (still needs another approval)

If there are additional unresolved aspects, we could add them as issues.

ninsch3000 (approved):

mfansler (approved):

Issue #13, opened 2021-05-19
D1: Specification - Compute Resources


mfansler:
Create specification for D1.

**Working Definition:** Obtain execution time and memory usage in MiB for method execution. If multiple rules or samples for method: sum time and get max MiB.

PR #93, opened 2021-06-02
docs: add parameter codes

### LINKED ISSUE #75: Compile parameter codes
[parameter_codes](https://github.com/iRNA-COSI/APAeval/blob/main/summary_workflows/parameter_codes.md) describing which datasets are to be run with which method parameters (if applicable) has to be created.

mfansler:
Resolves #75 

This is a working proposal. Please use review for discussion.

### DIFF
--- a/summary_workflows/parameter_codes.md
+++ b/summary_workflows/parameter_codes.md
@@ -1,23 +1,58 @@
 # Parameter codes
 
-## Input to summary workflows
+Parameter codes uniquely identify what combinations of input files and parameters were used to run an execution workflow. 
+These codes are used in summary workflows to determine appropriate comparison files.
 
-Input to summary workflows must be in the format of:
+## Formatting
 
-`PARAMCODE/METHOD/PARAMCODE_METHOD_OUTCODE.ext`
+Input to summary workflows must be in the following format:
 
-Example: AA/MISO/AA_MISO_01.bed
+```
+PARAMCODE/METHOD/PARAMCODE_METHOD_OUTCODE.ext
+```
 
-`PARAMCODE` is described here. For more information on naming scheme refer to the [execution workflow README][ex_readme]. 
-
-# Parameter code example table
+This document specifies the `PARAMCODE` component of this string. For more information on the full naming scheme refer to the [execution workflow README][ex_readme]. 
 
 
-| PARAM_CODE | dataset | param_setting| 
-| - | -| - |
-AA | P19 | cores=4 |
-AB | P19 | cores=12,mode=A |
-AC | P19_subset | cores=12,mode=A |
+# Working Parameter code table
+
+
+| PARAMCODE | dataset/samples                                                                                    | file name                         | matching ground truth                                                                              | run type | genome/annotation | explicit param\_settings | processing directive                             | benchmarks     |
+| --------- | -------------------------------------------------------------------------------------------------- | --------------------------------- | -------------------------------------------------------------------------------------------------- | -------- | ----------------- | ------------------------ | ------------------------------------------------ | -------------- |
+| AA        | SRR6795720                                                                                         | Mayr\_NB\_R1                      | SRR1005606<br>                                                                                     | indiv    | GRCh38            | 4 cores                  | individual sample                                | I2,I3,I4,Q2,Q3 |
+| AB        | SRR6795721                                                                                         | Mayr\_NB\_R2                      | SRR1005607<br>                                                                                     | indiv    | GRCh38            | 4 cores                  | individual sample                                | I2,I3,I4,Q2,Q3 |
+| AC        | SRR6795723                                                                                         | Mayr\_NB\_R3                      | SRR6795688<br>                                                                                     | indiv    | GRCh38            | 4 cores                  | individual sample                                | I2,I3,I4,Q2,Q3 |
+| AD        | SRR6795724                                                                                         | Mayr\_NB\_R4                      | SRR6795689                                                                                         | indiv    | GRCh38            | 4 cores                  | individual sample                                | I2,I3,I4,Q2,Q3 |
+| AE        | SRR6795718                                                                                         | Mayr\_CD5B\_R3                    | SRR6795684                                                                                         | indiv    | GRCh38            | 4 cores                  | individual sample                                | I2,I3,I4,Q2,Q3 |
+| AF        | SRR6795719                                                                                         | Mayr\_CD5B\_R4                    | SRR6795685                                                                                         | indiv    | GRCh38            | 4 cores                  | individual sample                                | I2,I3,I4,Q2,Q3 |
+| AG        | SRR6795726                                                                                         | Mayr\_M\_R2                       | SRR6795691                                                                                         | indiv    | GRCh38            | 4 cores                  | individual sample                                | I2,I3,I4,Q2,Q3 |
+| AH        | SRR6795713                                                                                         | Mayr\_GC\_R2                      | SRR6795693                                                                                         | indiv    | GRCh38            | 4 cores                  | individual sample                                | I2,I3,I4,Q2,Q3 |
+| AI        | SRR6795715                                                                                         | Mayr\_GC\_R1                      | SRR6795692                                                                                         | indiv    | GRCh38            | 4 cores                  | individual sample                                | I2,I3,I4,Q2,Q3 |
+| AJ        | SRR6795720, SRR6795721, SRR6795723, SRR6795724, SRR6795718, SRR6795719, SRR6795726, SRR6795713, SRR6795715 | Mayr\_allBcell\_group             | SRR1005606, SRR1005607, SRR6795688, SRR6795689, SRR6795684, SRR6795685, SRR6795691, SRR6795693, SRR6795692 | group    | GRCh38            | 4 cores                  | merge all samples                                | I2,I3,I4       |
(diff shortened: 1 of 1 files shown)


summary_workflows/parameter_codes.md
```diff
@@ -1,23 +1,58 @@
...
-AC | P19_subset | cores=12,mode=A |
+
+| PARAMCODE | dataset/samples                                                                                    | file name                         | matching ground truth                                                                              | run type | genome/annotation | explicit param\_settings | processing directive                             | benchmarks     |
+| --------- | -------------------------------------------------------------------------------------------------- | --------------------------------- | -------------------------------------------------------------------------------------------------- | -------- | ----------------- | ------------------------ | ------------------------------------------------ | -------------- |
+| AA        | SRR6795720                                                                                         | Mayr\_NB\_R1                      | SRR1005606<br>                                                                                     | indiv    | GRCh38            | 4 cores                  | individual sample                                | I2,I3,I4,Q2,Q3 |
```
uniqueg, line 22: This looks really nice! But I still think we should run all of those with 1, 4 and 16 cores. Perhaps we could append the cores to the parameter codes in the filenames so as not to explode the table (and then remove the column). So AA1, AA4, AA16 would be run AA with 1, 4, 16 cores. Having these data will allow us to evaluate the parallelization efficiency and scalability. And it should be easy to run these

mfansler: I like that idea. One thing I worry about is ending up with something like:

```
AJ1/MISO/AJ1_MISO_01.bed
AJ4/MISO/AJ4_MISO_01.bed
AJ16/MISO/AJ16_MISO_01.bed
```

Where, all those outputs should be identical. Which one(s) should say benchmark I2 assess?

The cores parameter should only matter for the I1 (#6), Q1 (#7), and D1 (#13) benchmarks (sidenote: we still need an output code for those). Perhaps a compromise would be that outputs that are agnostic to resources omit the cores count, so

```
AJ/MISO/AJ_MISO_01.bed
```

while the resource benchmarking outputs include one, like,

```
AJ4/MISO/AJ4_MISO_04.json
```

Perhaps that is even more confusing? 😬

dominikburri: I really like the table! I think it includes everything. 

As for the number of cores: It should just be a new parameter code. Most fields will be duplicated, but this shouldn't be a worry. It also gives the flexibility to adjust the benchmarks used etc. E.g. AP with one core will be just AQ and AP with 16 cores will be AR. 

To make it more readable, we could use the first letter in a more logical way. E.g. instead of "A" we could use the first letter of the file name, in this case "M".

Maybe I'm just overthinking the problem.

ninsch3000: I like it!
Remarks:
- when I suggested this format earlier, @uniqueg and @dominikburri were strongly opposing to individual samples having their own param code. Do you still want to address that issue?
- Let's keep the codes in concurrent alphabetical order, else it will get complicated and more dangerous to accidentally introduce duplicate codes
- Agree with @mfansler that appending cores to param codes will produce triplicate directories and identical output files. Where will the number of cores be specified in nextflow/snakemake? For Snakemake: if it was in the config, one could use it only to build the _outfile name for out 04_ like so:
```
# config.yaml
param_code: AJ
cores: 4

# Snakefile
out1 = os.pathjoin(config["param_code"], config["method_name]", "_".join(config["param_code]", config["method_name"], "01.bed"))
out4 = os.pathjoin(config["param_code"], config["method_name]", "_".join(config["param_code]", config["method_name"], "04", "config["cores"], ".json"))
```
- Adapting Mervins example: 
```
AJ/MISO/AJ_MISO_01.bed
AJ/MISO/AJ_MISO_04_4.json
```
Well, not sure... out 01-03 would then be overwritten at each new run with different cores. That ok? But might be confusing for executors. So maybe we should just keep the original outline

uniqueg: What I meant is that we wouldn't need a I1, Q1 & D1 anymore, but have compute performance monitored for _every_ benchmark (or, to use the OpenEBench terminology, make the compute performance _metrics_ of every _challenge_). This would give us more fine-grained insights into where individual tools struggle, performance-wise. It may not be all that interesting maybe, but it seems to me that keeping track of performance parameters for every run is something like "good practice"? Now, whether to run every tool _n_ times is debatable, guess we'd need to consider cost/benefit here. I'm totally fine with leaving that and doing it only once.

However, just to respond to some arguments:

As for producing _n_ times the data: guess that can't be avoided if we want to run the tools _n_ times, regardless of how we name the file.

As for choosing which one of many (supposedly identical) output files to use afterwards (for other benchmarks): we will need to arbitrarily define that (or have the tools define that for us by using defaults) regardless of whether we run it _n_ times or just once. And as for them being identical or not: this would already give us another metric we could _always_ check for, for every run: are the results reproducible (using different amounts of cores)? 

Yes, still against having parameter codes for individual samples, as those would then just be short names for samples and wouldn't reflect what is, in OEB parlance, a "challenge" (describing a specific set of inputs). To construct that we would then need to somehow merge different sample codes along some rules into a complex parameter/challenge code, and I think that's complicated. So, by all means, we should have _one_ simple code that identifies a specific way each tool should be run (and that should reasonably lead to identical or extremely similar results every time it is run for that tool). Of course, that doesn't mean that a set of inputs couldn't be just one sample, if that is what we really want to have for a specific run (I can't really judge that because I was following the metrics discussions at best peripherally).

I also agree with @ninsch3000 that it's best to keep the 2-letter code in alphabetical order, without any semantics, for the reasons she mentioned. But perhaps even better, we could follow the rules of the challenge IDs defined in the OpenEBench [schema definition](https://github.com/inab/benchmarking-data-model/blob/680cb6b28215a2ea9236c80ff906008134dee545/json-schemas/1.0.x/challenge.json), which call for the codes being generated in accordance with this regex: `^OEBX[0-9]{3}[A-Z0-9]{7}$` So that would be, e.g., `OEBX123ABCD0EF` or `OEBX00000000AA`, so basically we could keep our 2-letter codes, but could prefix them with `OEBX0000000`. Or we could use the first three digits to indicate the cores: `OEBX00100000AA`, `OEBX00400000AA`, `OEBX01600000AA`.

I'm not sure I understand the comment about how these params are specified in Snakemake or Nextflow. I mean, I know it is convenient if it could be named right in the workflows, but if not, surely post-processing could fix that. But did you mean there are limitations as to what we could do right inside Snakemake even _if_ we can access the number of cores as a parameter?

Anyway, let's vote (thumbs up or thumbs down) on a couple of things :)

uniqueg: Generate parameter codes in alphabetical order = thumbs up
Design semantic parameter codes = thumbs down

uniqueg: :heart: = Use 2-letter code
:rocket: = Use prefix to generate OEB challenge `_id`-compatible parameter codes

uniqueg: :heart: = run every challenge (i.e., set of inputs) _once_
:rocket: = run every challenge _multiple times_ to assess reproducibility and/or parallelization performance

uniqueg: :heart: = use different challenge codes for different # of CPUs or technical replicates; implies cross-_challenge_ comparison metrics (!)
:rocket: = codify # of CPUs and technical replicates in each challenge code

Suggestions for codifying:

* `OEBXnnn###00AA`, where `nnn` signifies the technical replicate, the `###` signifies the core number and `AA` signifies the set of inputs (example: `OEBX00301600AD`)
* `AA_nnn_###` (example: `AD_003_016`)

summary_workflows/parameter_codes.md
```diff
@@ -1,23 +1,58 @@
 # Parameter codes
 
-## Input to summary workflows
+Parameter codes uniquely identify what combinations of input files and parameters were used to run an execution workflow.
```
uniqueg, line 3: Considering the OpenEBench terminology, it may be better to call these _challenge codes_. What we _currently_ call challenges (identification, quantification, differential usage) would then become _benchmarking events_, and what we currently call _benchmarks_ would become _metrics_ (again?).

It's annoying, but we're unlikely to change the OpenEBench specifications/schemas, which have been around for a while and have some good thinking in them. And it would be good to adopt _one_ consistent terminology.

summary_workflows/parameter_codes.md
ninsch3000: Agree with all suggestions, but I still think we should have 1 param code per sample in *some cases* (As it is proposed in the table). More specifically, when we have a matching ground truth file for *each* sample and want to asses that, we wouldn't want the tools to merge several samples and give us only one output file for all samples. As we can't anticipate at the moment how merging or not merging of samples can be controlled for each tool, I think it is the most robust way to also have samples with their own param code. Another possibility might be to require the execution workflows to have a rule/directive that controls whether the tool receives a group of samples or individual samples as an input. This directive would then have to be configured according to the `run_type` or `processing_directive` of the param code table.    
That said, please **do note** that in other param codes a group of samples is specified, and those can be the *same* samples that earlier had their own param code. But now they would be allowed to be merged and will have to be compared to a merged ground truth. We never suggested to have *only* sample specific param codes.    
There are other datasets not included in this first param code table, where we do **not** have one ground truth per sample, and then of course we wouldn't run individual samples with their own param code but group them according to how their ground truths had been grouped.

PR #97, opened 2021-06-02
feat: barplot script for differential usage benchmarks

chilampoon:
## Checklist

- [x] I have tested my code

If we have results.

Some post-processings can see https://github.com/plger/diffUTR_paper/blob/main/figures/figures.Rmd

### DIFF
--- a/summary_workflows/differential_usage/diff.usage.barplot.py
+++ b/summary_workflows/differential_usage/diff.usage.barplot.py
@@ -5,10 +5,10 @@ import pandas as pd
 import matplotlib.pyplot as plt
 
 
-def bar_plot(df_path, plot_out):
+def bar_plot(files, plot_out):
 	'''
 	Inputs:
-		- df_path: path to a directory of tsv files containing 
+		- files: a list of tsv files containing 
 			  	gene_id & usage changes (e.g. deltaPPAU in QAPA);
 			  	NOTE that these genes have been filtered by FDR/q-value.
 			  	File names are like 'AA_diffUTR_03.tsv'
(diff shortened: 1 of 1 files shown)


chilampoon:
Example (I used the same two tsv files...)

image

uniqueg (dismissed):

summary_workflows/differential_usage/diff.usage.barplot.py
```diff
@@ -0,0 +1,64 @@
...
+	fig.savefig(plot_out)
+
+
+def main(args):
+	df_path = args['pathToTSV'][0] # directory to tsv files
+	out_path = args['out'][0] # directory to output image
+	plot_out = out_path + '/diff.usages.pdf'
+	#cutoff = float(args['cutoff'][0])
+
+	bar_plot(df_path,plot_out)
```
lschaerfen, line 53: Might be better to let the user specify the output file name instead of just the directory.

lschaerfen:
I don't fully understand why the thresholds are different and need to be hard coded for the different methods, but that might be on me. Looks good to me!

lschaerfen (approved):

uniqueg:
Actually, I think @lschaerfen brings up two good points. Perhaps, before we merge, you can replace out dir with out file and at least increase a little the documentation regarding the different cutoffs and why/how they need to be specified (i.e., how can I find out what to put here for a given set of methods)

chilampoon:
> I don't fully understand why the thresholds are different and need to be hard coded for the different methods, but that might be on me. Looks good to me!

Different methods output different statistics, if using _one cutoff_ here, we have to normalize the stats such as UTR length changes across all the methods first

summary_workflows/differential_usage/diff.usage.barplot.py
```diff
@@ -0,0 +1,63 @@
...
+	for file in files:
+		method = re.sub('_03.tsv$', '', os.path.basename(file))
+		method = re.sub('^.*_', '', method)
+
+		# hardcode the cutoffs since each method has its own calculation
+		cutoff = 0
+		if method == "QAPA":
+			cutoff = 20
+		elif method == "diffUTR":
+			cutoff = 0.5 #not sure, can ask the author
```
yuukiiwa, line 31: I agree with the points @lschaerfen and @uniqueg mentioned. I personally prefer flexibility in specifying the cutoffs unless the authors specifically mention that a certain threshold is a gold standard, which requires referencing. Other than that, everything looks good to me!

yuukiiwa (approved):
I agree with the points @lschaerfen and @uniqueg mentioned. I personally prefer flexibility in specifying the cutoffs unless the authors specifically mention that a certain threshold is a gold standard, which requires referencing. Other than that, everything looks good to me!
Sorry for double commenting

chilampoon:
Yea should be flexible
If APAeval is continued after today's presentation, I may write a script for post-processing for results from various methods.

> I agree with the points @lschaerfen and @uniqueg mentioned. I personally prefer flexibility in specifying the cutoffs unless the authors specifically mention that a certain threshold is a gold standard, which requires referencing. Other than that, everything looks good to me! We can just have it stored here first and see how we should integrate the visualization codes into the `consolidation` step.

PR #109, opened 2021-06-03
feat: add OpenEBench challenge/metric schemas for Q2

dominikburri:


openebench/APAeval_schemas/Challenges/APAeval_challenge1.json
```diff
@@ -0,0 +1,32 @@
...
+        "Mervin.M.Fansler"
+    ],
+    "metrics_categories": [
+        {
+            "category": "assessment",
+            "description": "Metrics used to benchmark the correlation between RNAseq-based poly(A) site quantification and poly(A)-sites quantification based on orthogonal 3'end seq data.",
+            "metrics": [
+                {
+                    "metrics_id": "APAeval:corr_coeff",
+                    "tool_id": "APAeval:Tool_QAPA.json"
```
daneckaw, line 23: If the examples in the repo below are correct, then tool_id would refer to a tool computing the metric, and a challenge would be comparing one or more metrics for a given dataset, but I don't know if it's a fixed rule or we have some freedom here.  
https://github.com/inab/benchmarking-data-model/tree/master/prototype-data/1.0.x/TCGA/Challenges

dominikburri: Yes, I think tool_id should point to a summary workflow, like the one we have now that computes the correlations. I will try to work on this and to improve, but not sure how far I get.

dominikburri:
Some comments on the last commits:

- Added the challenge1, which corresponds to our Q2 benchmark with the correlation value.
- Added description of correlation and aggregation tool. Based on the examples, they are Docker images on a github page. 
  - I don't know how this integrates with nextflow-tower, but I think the correlation tool is the summary workflow we have, resp. one task of it.
- I created two metric events that execute QAPA and LABRAT on Q2 challenge. This needs the input data, here the output of the execution workflows and the ground truth data. This step will create a new dataset that will be used for the following aggregation.
- The aggregation event now gathers the metrics of QAPA and LABRAT (in this case the correlation values) and gathers it with the tool aggregate to one dataset. 
- This dataset includes all correlation values is now used for visualisation. This step is still missing. 

Let me know if I understood something wrong and the described procedure is not correct.

daneckaw:
If I understood correctly, the schemas for summary workflow and possibly for aggregation/consolidation workflow are already created and added to the database, so we'll have to update the challenge schema with correct tool_id's when we get them tomorrow.

openebench/APAeval_schemas/Challenges/APAeval_challenge1.json
```diff
@@ -0,0 +1,38 @@
+{
+    "_id": "APAeval:2021-06-04_Q2",
+    "_schema": "https://www.elixir-europe.org/excelerate/WP2/json-schemas/1.0/Challenge",
+    "name": "Correlation benchmark for RNA-seq poly(A) site quantification.",
+    "acronym": "Q2",
+    "benchmarking_event_id": "APAeval:2021-06-04",
```
uniqueg, line 6: Shouldn't this be: OEBE0070000000?

openebench/APAeval_schemas/Challenges/APAeval_challenge1.json
```diff
@@ -0,0 +1,38 @@
...
+    "is_automated": true,
```
uniqueg, line 7: Shouldn't this be `false`?

openebench/APAeval_schemas/aggregation_events/aggregationevent_Q2.json
```diff
@@ -0,0 +1,29 @@
...
+    "action_type": "AggregationEvent",
+    "challenge_id": "APAeval:2021-06-04",
+    "dates": {
+        "creation": "2021-06-04T00:00:00Z",
+        "reception": "2021-06-04T00:00:00Z"
+    },
+    "tool_id": "APAeval:aggregation",
+    "involved_datasets": [
+        {
+            "dataset_id": "APAeval:2021-06-04_AA_QAPA_Q2_01_corr",
```
dominikburri, line 13: This should be replaced by the actual `dataset_id`.

openebench/APAeval_schemas/Challenges/APAeval_challenge1.json
lauportell: yes!

openebench/APAeval_schemas/Challenges/APAeval_challenge1.json
lauportell: for the moment doesn't matter, this is only a flag for us

openebench/APAeval_schemas/aggregation_events/aggregationevent_Q2.json
dominikburri: Same for the other datasets in this file.

openebench/APAeval_schemas/metrics_events/metricsevent_AA_LABRAT_Q2.json
```diff
@@ -0,0 +1,29 @@
...
+        "reception": "2021-06-04T00:00:00Z"
+    },
+    "tool_id": "APAeval:Tool_correlation",
+    "involved_datasets": [
+        {
+            "dataset_id": "APAeval:2021-06-04_AA_LABRAT_Q2_01",
+            "role": "incoming"
+        },
+        {
+            "dataset_id": "APAeval:2021-06-04_<ground_truth>",
```
dominikburri, line 17: Again, replacement with actual IDs, here for the ground truth BED files

openebench/APAeval_schemas/aggregation_events/aggregationevent_Q2.json
lauportell: have a look at any of these examples to check what are the fields that are mandatory:
https://github.com/inab/benchmarking-data-model/tree/master/prototype-data/1.0.x/QfO/Challenges
dataset_id is not necessary I am afraid

uniqueg: Not in the challenge schema, in the testAction schema it is. Will leave this for now, as apparently we won't need the testAction schemas

uniqueg (approved):
Tested with https://github.com/fairtracks/fairtracks_validator, remaining issues should be resolved once on OpenEBench. Exception: testAction schemas will contain unavaialble dataset ids, but these schemas won't be used for now

uniqueg (approved):
Added my fix on the challenge IDs (had pushed it to a wrong branch earlier)

PR #117, opened 2021-06-03
fix: account for strand in QAPA post-processing

mrgazzara:
`ln[8]` and `ln[9]` correspond to "UTR3.Start" and "UTR3.End", respectively. These are genomic coordinates and so depending on the strand the actual PAS coordinate will change to being either the end of the transcript (`if strand=="+"` we need the "UTR3.End" / `ln[8]`. Otherwise we assume `strand =="-"` then we need "UTR3.Start" / `ln[8]`)

Fixes #{ISSUE_NUMBER}

## Checklist

- [ ] I have tested my code

### DIFF
--- a/execution_workflows/QAPA/bin/make_quant_bed.py
+++ b/execution_workflows/QAPA/bin/make_quant_bed.py
@@ -9,6 +9,13 @@ header_fields=['chrom','chromStart','chromEnd','name','score','strand']
 qapa_quant_bed.write('\t'.join(header_fields)+'\n')
 for ln in qapa_results_file:
  ln=ln.strip().split('\t')
- fields=[ln[4],ln[8],ln[9],ln[0],ln[13],ln[7]]
+ strand = ln[7] 
+ #if "+" strand gene we want to grab "UTR3.End" as PAS
+ if strand == '+': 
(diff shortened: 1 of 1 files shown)


uniqueg (dismissed):
Guess it makes sense, and code looks good. @yuukiiwa?

execution_workflows/QAPA/bin/make_quant_bed.py
```diff
@@ -9,6 +9,13 @@
 qapa_quant_bed.write('\t'.join(header_fields)+'\n')
```
yuukiiwa, line 9: Can you remove this line, please? Thank you!

yuukiiwa:
The changes on the strands look good! I think you can also fix the PAS bug within this PR, which I am not exactly sure how to fix it. Thank you very much!

execution_workflows/QAPA/bin/make_quant_bed.py
mrgazzara: Removed this and leading "chr" in commit 3b8b138499212e8c24d009d3ff83638c2ce4ca00

dominikburri:
Fixed in #196

Issue #62, opened 2021-05-28
Create Utils channel


dominikburri:
From the first progress report discussion we want to create a place for common tasks.

The idea is that some tasks are reoccurring, like extracting polyA sites from a database, and should not be scattered around in individual execution and/or summary workflows. It therefore makes sense to create a directory that gathers such tasks. 

The following steps should be considered

- [ ] Create a project page to gather related issues.
- [ ] Create a directory `utils`.
- [ ] Write a `README.md` within this directory describing the 
  - [ ] purpose,
  - [ ] usage (How-to) and
  - [ ] description of implemented scripts. 
- [ ] Append to the main `README.md` that 
  - [ ]  When creating a new workflow, the utils directory should be checked for existing common tasks.
  - [ ]  If some common task is not yet in utils, a github issue should be created. Then it should be written and added as individual script.

Additionally, the following should be considered when implementing the structure:

- When an execution or summary workflow is run, it should be able to use the scripts within the utils directory. 
- The individual scripts should be unit tested before being used in the workflows. That is, it should be ensured that the script is flexible, robust and actually doing what is intended to.
- ...

PR #151, opened 2021-07-15
feat: execution workflow APAtrap

### LINKED ISSUE #29: Execution Workflow: APAtrap
## WHAT    
**Write** execution workflow for [APAtrap](https://sourceforge.net/projects/apatrap/). Use the provided [small files for testing](https://github.com/iRNA-COSI/APAeval/tree/main/tests/test_data) (running the workflow on real data is a different issue).

## CHECKLIST   
* [x] Use [snakemake template](https://github.com/iRNA-COSI/APAeval/tree/main/docs/templates/snakemake) or [nextflow template](https://github.com/iRNA-COSI/APAeval/tree/main/docs/templates/nextflow) to create your workflow.
* [x] Comment your code
* [x] Run individual rules/processes in either conda envs or docker/singularity containers for reproducibility
* [x] Input:  .bam or .fastq from [test_data]((https://github.com/iRNA-COSI/APAeval/tree/main/tests/test_data))

* [x] Output: Adhere to output specification for Identification challenge
Fields:

* [x] Output: Adhere to output specification for quantification challenge

* [x] Output: Adhere to output specification for differential usage challenge

    > - gene ID

faricazjj:
Fixes #29 

## Checklist

- [ ] I have tested my code

### DIFF
--- a/execution_workflows/APAtrap/main.nf
+++ b/execution_workflows/APAtrap/main.nf
@@ -1,0 +1,48 @@
+#!/usr/bin/env nextflow
+/*
+========================================================================================
+                         PILOT BENCHMARK PIPELINE RUNNING QAPA
+========================================================================================
+*/
+
+nextflow.enable.dsl = 2
+
+////////////////////////////////////////////////////
+/* --               PRINT HELP                 -- */
+////////////////////////////////////////////////////
+
+if (params.help) {
+    //print something here
+    exit 0
+}
+
+////////////////////////////////////////////////////
+/* --          PARAMETER CHECKS                -- */
+////////////////////////////////////////////////////
+
+// Check that conda channels are set-up correctly
+//if (params.enable_conda) {
+//    Checks.check_conda_channels(log)
+//}
+
+// Check AWS batch settings
+//Checks.aws_batch(workflow, params)
+
+// Check the hostnames against configured profiles
+//Checks.hostname(workflow, params, log)
+
+////////////////////////////////////////////////////
+/* --           RUN MAIN WORKFLOW              -- */
+////////////////////////////////////////////////////
+
+workflow {
+    /*
+     * RUN APATRAP WORKFLOW
+     */
+    include { EXECUTE_APATRAP } from './workflow/execute_apatrap' addParams( [:] )
+    EXECUTE_APATRAP ()
+}
+
+////////////////////////////////////////////////////
+/* --                  THE END                 -- */
+////////////////////////////////////////////////////
--- a/execution_workflows/APAtrap/modules/preprocessing.nf
+++ b/execution_workflows/APAtrap/modules/preprocessing.nf
@@ -1,0 +1,27 @@
+// Import generic module functions
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+params.options = [:]
+def options    = initOptions(params.options)
+
+/*
+    Convert the input .bam file to .bedgraph for APAtrap
+*/
+process PREPROCESSING {
+
+    publishDir "${params.outdir}/apatrap", mode: params.publish_dir_mode
+    container "faricazjj/apatrap"
+
+    input:
+    tuple path(bam_file), path(bai_file)
+
+    output:
+    path "$input_bedgraph", emit: ch_3utr_input
+
+    script:
+    input_bedgraph = "3utr_input.bedgraph"
+    """
+    bamCoverage -b $bam_file -o $input_bedgraph
+    """
+
+}
--- a/execution_workflows/APAtrap/subworkflows/input_check.nf
+++ b/execution_workflows/APAtrap/subworkflows/input_check.nf
@@ -1,0 +1,22 @@
+/*
+ * Check input samplesheet and get read channels
+ */
+
+params.options = [:]
+
+include { CHECK_SAMPLESHEET;
+          get_sample_info } from '../modules/check_samplesheet' addParams( options: params.options )
+
+workflow INPUT_CHECK {
+    take:
+    samplesheet // file: /path/to/samplesheet.csv
+    
+    main:
+    CHECK_SAMPLESHEET ( samplesheet )
+        .splitCsv ( header:true, sep:',' )
+        .map { get_sample_info(it) }
+        .set { ch_sample }
+
+    emit:
+    ch_sample // [ sample, barcode, fasta, gtf, is_transcripts, annotation_str ]
+}
--- a/execution_workflows/APAtrap/subworkflows/run_apatrap.nf
+++ b/execution_workflows/APAtrap/subworkflows/run_apatrap.nf
@@ -1,0 +1,28 @@
+/*
+ * Run APAtrap
+ */
+
+include { PREPROCESSING } from '../modules/preprocessing' addParams( options: [:] )
+
+workflow RUN_APATRAP {
+    take:
+    ch_sample
+
+    main:
+    ch_sample
+        .map { it -> [ it[3], it[4] ] }
+        .unique()
+        .set { ch_preprocess }
+
+    /*
+        Preprocess the input file: convert from bam to bedgraph
+    */
+    PREPROCESSING( ch_preprocess )
+
+    /*
+        Perform first step of APAtrap: identify distal 3'utr
+    */
+    ch_identify_3utr_input = PREPROCESSING.out.ch_preprocessing_output
+    IDENTIFY_DISTAL_3UTR( ch_identify_3utr_input )
+}
+


execution_workflows/APAtrap/modules/identify_distal_3utr.nf
```diff
@@ -0,0 +1,26 @@
...
+        input:
+        tuple path(genome_file), path(reads_bedgraph_file)
+
+        output:
+        path "$utr_output", emit: ch_predictapa_input
+
+        script:
+        utr_output = "predictapa_input.bed"
+        """
+        identifyDistal3UTR -i $reads_bedgraph_file -m $genome_file -o $utr_output
```
SamBryce-Smith, line 24: Looking at docs and source code there are a number of optional parameters for `identifyDistal3UTR`. I think these should be specified in pipeline (and set to default values in a config file) so we have the option to modify if we wish.

Optional parameters:
```
-w  window size used to scan the mapping result, default is 100 (has to be at least 20).
-e  pre-extension size of each 3'UTR, default is 10000 (has to be at least 100)
-c  minimum coverage of the end of the distal 3'UTR with comparing to the whole transcript, default is 0.05 (valid range 0-1).
-p  minimum percentage of valid nucleotides in a scanning-window, default is 0.8 (valid range 0-1).
-s  gene symbol file.
```
I don't think specifying gene symbol file is necessary though, but all the other optional parameters should be defined in my opinion.

execution_workflows/APAtrap/modules/de_apa.nf
```diff
@@ -0,0 +1,29 @@
...
+        path "$de_apa_output", emit: ch_de_apa_output
+
+        script:
+        de_apa_output = "de_apa_output.txt"
+        """
+        #!/usr/bin/Rscript
+
+        library(deAPA)
+        deAPA('$de_apa_input', '$de_apa_output', 1, 1, 1, 1, 20)
+        """
```
SamBryce-Smith, line 28: I don't think parameters should be hardcoded in the pipeline.  They should be defined and described in a config file (they can be left at default values for now though).

The 3rd & 4th arguments of the `deAPA` function are group1 & group2 respectively according to docs. Does setting both to 1 (i.e. comparing samples in group1 to samples in group1) produce valid comparisons *between* groups?

execution_workflows/APAtrap/bin/postprocessing.py
```diff
@@ -0,0 +1,67 @@
...
+def reformat_bed(file_in):
+    """
+    This function reformats the txt deAPA output file to files for
+    identification, quantification, and differential challenges
+    :param file_in: txt file to be reformatted
+    :return: N/A
+    """
+    identification_out = open("apatrap_identification_output.bed", "wt")
+    quantification_out = open("apatrap_quantification_output.bed", "wt")
+    differential_out = open("apatrap_differential_output.tsv", "wt")
```
SamBryce-Smith, line 25: I don't think output filenames should be hardcoded. Would be better to define output file name when call script at command line and have option to specify output filename in config file

execution_workflows/APAtrap/README.md
```diff
@@ -0,0 +1,56 @@
+# APAeval nextflow pilot run
```
SamBryce-Smith, line 1: This line can be deleted

execution_workflows/APAtrap/README.md
```diff
@@ -0,0 +1,56 @@
...
+The [paper](https://academic.oup.com/bioinformatics/article/34/11/1841/4816794) is titled APAtrap: identification and quantification of 
+alternative polyadenylation sites from RNA-seq data
+The [application download](https://sourceforge.net/projects/apatrap/files/) is free to download
+and [user manual](https://sourceforge.net/p/apatrap/wiki/User%20Manual/) was used as a referrence
+to create the nextflow pipeline flow of this module
+
+### Steps to run this:
+ - Replace "path_to" in samplesheet_example_files.csv with the path to the reference input files
+ - Check the path to the input files with `pwd` and replace the `path_to` in samplesheet_example_files.csv with the path from the `pwd` command
+ - Pull the required docker image by doing `docker pull docker.io/faricazjj/apatrap:latest`
```
SamBryce-Smith, line 19: update docker link to `docker.io/apaeval/apatrap:latest` as in rest of pipeline

execution_workflows/APAtrap/modules/predict_apa.nf
```diff
@@ -0,0 +1,26 @@
...
+        input:
+        tuple path(reads_bedgraph_file), path(predict_apa_input)
+
+        output:
+        path "$predict_apa_output", emit: ch_de_apa_input
+
+        script:
+        predict_apa_output = "de_apa_input.txt"
+        """
+        predictAPA -i $reads_bedgraph_file -g 1 -n 1 -u $predict_apa_input -o $predict_apa_output
```
SamBryce-Smith, line 24: Again I don't think parameter values should be hardcoded. For differential usage benchmarks we will definitely have more than one group (`-g 1`) and > 1 sample per group (`-n 1`). As with `identifyDistal3UTR` I think optional parameters should be specified in call and defined in a config file, even if they are set to defaults for now.

```
-d  minimum degree of coverage variation between two adjcent APA sites, >0 and <1, default is 0.2.
-c  minimum average coverage required for each 3'UTR, >=10, default is 20.
-a  minimum distance between the predicted APA sites, >=20, default is 100.
-w  window size used to scan the profile, >=20, default is 50.
```

SamBryce-Smith (commented):
Great work @faricazjj! Generally my main issues are with parameter hardcoding. I think as a general rule parameters should be customisable and defined in a config file, even if they are initially set to the default values.

Will have another look tomorrow, I haven't had a chance to try following your instructions and running the pipeline on test data but want to record what I've found so far

I have no experience with Nextflow so will leave comments on that front to the more experienced reviewers

faricazjj:
> Great work @faricazjj! Generally my main issues are with parameter hardcoding. I think as a general rule parameters should be customisable and defined in a config file, even if they are initially set to the default values.
> 
> Will have another look tomorrow, I haven't had a chance to try following your instructions and running the pipeline on test data but want to record what I've found so far
> 
> I have no experience with Nextflow so will leave comments on that front to the more experienced reviewers

Thanks for reviewing @SamBryce-Smith ! I'll address your comments and push out the changes soon 😁

execution_workflows/APAtrap/main.nf
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env nextflow
+/*
+========================================================================================
+                         PILOT BENCHMARK PIPELINE RUNNING QAPA
```
faricazjj, line 4: will update the comment to 'RUNNING APATRAP'

execution_workflows/APAtrap/README.md
faricazjj: deleting this line since it's not necessary to pull the required docker image--the image is automatically pulled when nextflow runs

execution_workflows/APAtrap/modules/de_apa.nf
faricazjj: @SamBryce-Smith  good point, thank you! Do you happen to know if we'll always provide two sample files for differential challenge? And if so, will these two sample files used for differential challenge also used for the other two challenges? Wondering since the samplesheet only has one pair of bam and bai files, but if we need two sample files for differential challenge then it seems like I'd have to change the samplesheet headers so that it allows for two bam and bai files so that the deAPA step would take in two samples  to produce valid comparisons between groups

SamBryce-Smith: I'm not sure about whether it would be two samples for each group in the differential challenge. I would expect it to be more, but our workflows should be able to count how many samples per group from the sample sheet (if it is formatted correctly).

I'm pretty sure the samples used for the differential challenge will be different to those used for the other challenges though.

RE your sample sheet, I will make comment again on review later (about to finish for day today) but I think it could be structured more as a 'long' rather than 'wide' table. Actually how it is at the moment gives the incorrect relationships about the files:

samplesheet_example_files.csv as of 0c9e44b :

the files in bam1 & bam2 are actually two **distinct** samples (control & siRNA KD), but this table would suggest they are the same sample 'test'

| sample | bam1                            | bai1                                | bam2                          | bai2                              | bed                              |
|--------|---------------------------------|-------------------------------------|-------------------------------|-----------------------------------|----------------------------------|
| test   | path_to/siControl_R1_2genes.bam | path_to/siControl_R1_2genes.bam.bai | path_to/siSrsf3_R1_2genes.bam | path_to/siSrsf3_R1_2genes.bam.bai | path_to/vm18.mouse.genemodel.bed |


If this table was 'longer', you could keep the same template/columns and just add rows if you have > 2 samples. Then you could also easily annotate which group they belong to as well.

My suggestion would be something like below. I'd also remove the `bed` column and place it in a more general config (i.e. global parameter for this given run), as I think we'll want to process each sample with the same annotation.

| sample    | group   | bam                             | bai                                 |
|-----------|---------|---------------------------------|-------------------------------------|
| siCTL_1   | siCTL   | path_to/siControl_R1_2genes.bam | path_to/siControl_R1_2genes.bam.bai |
| siSrsf3_1 | siSrsf3 | path_to/siSrsf3_R1_2genes.bam   | path_to/siSrsf3_R1_2genes.bam.bai   |

Hopefully that makes sense, from personal experience I find it easier to deal with 'longer' tables anyway. I suspect you may have to rejig how you input info from sample table to pipeline but I think the extra flexibility is a bonus. Happy to provide suggestions/discuss if you feel you need it (though I'm out of my depth with nextflow) :)

P.S. - we have the test data in the main repository. If the example sample sheet could contain the relative paths to the test data already, that would make it that bit easier for us to do a test run of the pipeline!

ninsch3000: Agree with Sam, it would be awesome if the sample sheet could already point to the test data in the APAeval repo (Although I just tried it but couldn't make nextflow accept relative paths. Is that a reality?).    

Also, I agree that the genemodel file could be specified somewhere else (not in the sample sheet), as it is the same for all samples (and sample sheet should ideally only contain sample specific information).    

Speaking of the genemodel file, I'm not comfortable with downloading it from UCSC and adding "chr" to the bedgraph files, as this could make the outputs incomparable (well, not *really* incomparable, but it's still a bit dirty) to outputs from other tools that use ensembl annotations. We should generate the genemodel in pre-processing (best to add to utils, see #62 )

SamBryce-Smith (commented):
Again ran out of time to get a successful test run but more comments inbound :) I haven't got around to downloading the BED12 file but everything worked up to that point.

Addressed my previous comments very well so this feels a lot like just tidying up now! Some general comments that I couldn't get into in other comments:

- Is there an example sample-sheet/instructions to run the differential usage workflow with our test data? If so, I think having a separate subheader for running differential_usage / identification & quantification workflows would be great

execution_workflows/APAtrap/modules/de_apa.nf
SamBryce-Smith: I had the same problem with relative paths too. Adding the absolute path from `pwd` before the relative path worked for me though but felt a bit dirty (workflow was able to generate bedgraphs at least).

ninsch3000:
Thanks so much for the work @faricazjj !!

My comments:

As I wrote above, I'm not comfortable with downloading the models and adding "chr", I guess it would be good to discuss that in the APAeval meeting and find a common solution. Actually, we could try creating the genemodel file ourselves as described in the APAtrap manual (https://sourceforge.net/p/apatrap/wiki/User%20Manual/), and this might already render the adding of "chr" unnecessary...

Unfortunately, I have no idea about nextflow, so can't say much there. But I ran the workflow on my laptop and it ran through :p

I'm just wondering, how does APAtrap know which samples to group if there are more than two samples? As far as I could see, the sample sheet doesn't contain this information...

execution_workflows/APAtrap/modules/identify_distal_3utr.nf
```diff
@@ -0,0 +1,67 @@
...
+                sample_files+="\$file "
+            done
+            identifyDistal3UTR -i \$sample_files \
+                -m $genome_file \
+                -o $utr_output \
+                -w $window_size \
+                -e $extension_size \
+                -c $min_coverage \
+                -p $min_percentage
+            """
```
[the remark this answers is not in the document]
ninsch3000: No clue, but maybe something like this: https://nextflow-io.github.io/patterns/index.html#_process_per_file_output

faricazjj:
> Thanks so much for the work @faricazjj !!
> 
> My comments:
> 
> As I wrote above, I'm not comfortable with downloading the models and adding "chr", I guess it would be good to discuss that in the APAeval meeting and find a common solution. Actually, we could try creating the genemodel file ourselves as described in the APAtrap manual (https://sourceforge.net/p/apatrap/wiki/User%20Manual/), and this might already render the adding of "chr" unnecessary...
> 
> Unfortunately, I have no idea about nextflow, so can't say much there. But I ran the workflow on my laptop and it ran through :p
> 
> I'm just wondering, how does APAtrap know which samples to group if there are more than two samples? As far as I could see, the sample sheet doesn't contain this information...

Thanks for the input @ninsch3000 ! :D To address your comments, the gene model file downloaded from UCSC is not edited in this workflow. The file that is being checked if the prefix 'chr' exists is the converted bedgraph from the input bam file. This is because without the prefix 'chr' in the bedgraph file that is used as the inputs for APAtrap, an error would occur and we can't run APAtrap. I think having the 'chr' prefix in the chromosome column of is pretty common but I'm not sure why we removed it from our test files :D

As for the grouping of samples, I clarified with the author and he said that we should put input files by group and in the some order of values specified by parameter -n. So for example, if we have two replicates for each condition, we'll then do
```
predictAPA -i condition1_replicate1.bedgraph condition1_replicate2.bedgraph condition2_replicate1.bedgraph condition2_replicate2.bedgraph -g 2 -n 2 2 -u hg19.utr.bed -o output.txt
```
I'll add this explanation in the README too and update the workflow, thanks for bringing it up you guys! @SamBryce-Smith @ninsch3000

execution_workflows/APAtrap/modules/identify_distal_3utr.nf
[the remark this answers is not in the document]
faricazjj: @SamBryce-Smith @ninsch3000 Are you guys referring do line 40-44 where we're concatenating the list of samples? The reason I'm doing this is exactly because nextflow processes things per file output (that CJ linked), but we don't want this since we need all the file outputs/sample files together in one identifyDistal3UTR call. Hence the bash script file name concatenation 😆  So yeah I'm not sure if nextflow can do this (couldn't find anything online) since it's innately designed to process per file output

execution_workflows/APAtrap/README.md
```diff
@@ -0,0 +1,68 @@
...
+columns:
+
+- sample: name of the sample
+- bam: BAM input file for the sample 
+- bai: BAI index file for sample's bam input
+- bed: gene model file in bed format. Can be obtained from [UCSC table browser](http://genome.ucsc.edu/cgi-bin/hgTables?hgsid=1133780495_lZCAEdlwBd7HbE03thrN4Tsi6lSF&clade=mammal&org=Mouse&db=mm10&hgta_group=genes&hgta_track=wgEncodeGencodeVM18&hgta_table=0&hgta_regionType=genome&position=chr12%3A56%2C694%2C976-56%2C714%2C605&hgta_outputType=bed&hgta_outFileName=)
+
+Each row in the samplesheet has the files needed for each sample file.
+When run_differential is set to false, at least one row is required to be provided.
+When run_differential is set to true, at least two rows (i.e. two conditions) are required to be provided.
```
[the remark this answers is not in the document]
faricazjj: Thanks for pointing it out @SamBryce-Smith ! Really good point!  I clarified with the author and he said that we should put input files by group and in the some order of values specified by parameter -n. So for example, if we have two replicates for each condition, we'll then do

predictAPA -i condition1_replicate1.bedgraph condition1_replicate2.bedgraph condition2_replicate1.bedgraph condition2_replicate2.bedgraph -g 2 -n 2 2 -u hg19.utr.bed -o output.txt
 
This means that the order of files in the samplesheet has no restrictions but I'll probably have to update the workflow such that it groups sample files by condition and then concatenates them next to each other

execution_workflows/APAtrap/conf/modules.config
```diff
@@ -0,0 +1,81 @@
...
+        }
+        'predict_apa' {
+            // Optional and necessary parameters for step 2 of APAtrap: predict apa
+
+            // Necessary paramters
+            // number of groups (treatments/conditions) of the input files, e.g. -g 2.
+            g = 1
+            // number of files(biological replicates) in each group (treatment/condition), e.g. -n 1 1.
+            // if there is only one file and one group, set n = 1
+            n = '1'
```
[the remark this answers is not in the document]
faricazjj: Copy pasting the reply on the readme here since the question is around the same issue :D 

Really good point! I clarified with the author and he said that we should put input files by group and in the some order of values specified by parameter -n. So for example, if we have two replicates for each condition, we'll then do

predictAPA -i condition1_replicate1.bedgraph condition1_replicate2.bedgraph condition2_replicate1.bedgraph condition2_replicate2.bedgraph -g 2 -n 2 2 -u hg19.utr.bed -o output.txt

This means that the order of files in the samplesheet has no restrictions but I'll probably have to update the workflow such that it groups sample files by condition and then concatenates them next to each other

execution_workflows/APAtrap/more_samplesheets/samplesheet_paired.csv
```diff
@@ -0,0 +1,2 @@
+sample,fastq1,fastq2,bam,bai,gff,fasta,bed,mart_export
+test,path_to/SRR3067957_1.fastq,path_to/SRR3067957_2.fastq,path_to/Aligned.out.sorted.bam,path_to/Aligned.out.sorted.bam.bai,path_to/gencode.vM26.primary_assembly.annotation.gff3,path_to/mm39.fa,path_to/qapa_3utrs.gencode_VM22.mm10.bed,path_to/mart_export.txt
```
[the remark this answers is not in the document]
faricazjj: oh yes thank you! This was from the template I copied over, wasn't sure if this file is needed for benchmarking. @yuukiiwa I saw that you uploaded the template! :D Was this file added for testing and thus safe to delete if we're not using it for ours or should we keep it and edit the header such that it's the same as our main samplesheet used to run the workflow?

execution_workflows/APAtrap/samplesheet_example_files.csv
```diff
@@ -0,0 +1,3 @@
+sample,bam,bai,bed
+Control,path_to/siControl_R1_2genes.bam,path_to/siControl_R1_2genes.bam.bai,path_to/vm18.mouse.genemodel.bed
+Srsf3,path_to/siSrsf3_R1_2genes.bam,path_to/siSrsf3_R1_2genes.bam.bai,path_to/vm18.mouse.genemodel.bed
```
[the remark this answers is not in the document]
faricazjj: ohh sorry I don't think I understand completely! When I developed this locally I thought I had provide absolute paths to these test files. I went to the folder with these files, run pwd, and then replace path_to with the path shown by pwd. If we have path_to/../../tests/test_data/, it looks like we're going to the right folder, and then we go out of it, and go back into test_data, unless I'm interpreting this incorrectly 😆 What is the ../../tests/test_data for?

execution_workflows/APAtrap/README.md
```diff
@@ -0,0 +1,68 @@
...
+and [user manual](https://sourceforge.net/p/apatrap/wiki/User%20Manual/) was used as a referrence
+to create the nextflow pipeline flow of this module
+
+### Steps to run this:
+ - We can either run the workflow to obtain identification and quantification outputs, or differential output. There
+   is currently no way to run the workflow once to get outputs for all three challenges
+ - samplesheet_example_files.csv contains paths to all the input files provided for each workflow run
+ - Replace "path_to" in samplesheet_example_files.csv with the path to the reference input files
+ - Check the path to the input files with `pwd` and replace the `path_to` in samplesheet_example_files.csv with the 
+   path from the `pwd` command
```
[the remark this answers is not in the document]
faricazjj: Yes you're right! I'll remove line 18 :)

execution_workflows/APAtrap/README.md
```diff
@@ -0,0 +1,68 @@
...
+Each APAtrap run results in files of the outputs of the challenges located under APAtrap/results/apatrap/challenges_outputs folder.
+For identification and quantification outputs, the files have sample names as prefixes to differentiate the different runs.
+The differential output file will stay as the name specified in modules.config file.
+
+
+## Notes
+The "chr" in the chromosome column in the input bam file needs to be there. Otherwise, identifyDistal3UTR will
+product an empty output file. The check for leading 'chr' in the chromosome column is done under 
+PREPROCESSING step. If no leading chr is detected, it'll be added.
+
```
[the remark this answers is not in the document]
faricazjj: Great idea, will do!

execution_workflows/APAtrap/samplesheet_example_files.csv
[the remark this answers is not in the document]
SamBryce-Smith: > I went to the folder with these files


Oooh sorry this is my bad, this is what I didn't do! I was sitting in the APAtrap directory, typed `pwd` and used that to replace `path_to/`, which I obviously realised wouldn't work as we don't have the test data in this directory. I thought it would be better to just use the relative paths so I tested that first, not knowing that Nextflow struggles with relative paths. My workaround works like this

```
# relative path to file is ../../tests/test_data/siControl_R1_2genes.bam
# in sample sheet I had path_to/../../tests/test_data/siControl_R1_2genes.bam
$ pwd
/home/APAeval/execution_workflows/APAtrap

# replace path_to with `/home/APAeval/execution_workflows/APAtrap`
$ ls /home/APAeval/execution_workflows/APAtrap/../../tests/test_data/siControl_R1_2genes.bam
 /home/APAeval/execution_workflows/APAtrap/../../tests/test_data/siControl_R1_2genes.bam
```

Both are equally valid in my opinion:

1. Keep as is, but specify in README to change to the `tests/test_data` repo and type `pwd` (we know where it is but others may not).
2. Go with my suggestion, and specify that you should be in `execution_workflows/APAtrap` when you type `pwd`

Slight preference to 2nd option as I'd generally assume people wanting to run this workflow would already be in the APAtrap directory and this saves jumping about, but probably my personal bias creeping in :P

@ninsch3000 @faricazjj thoughts?

execution_workflows/APAtrap/modules/identify_distal_3utr.nf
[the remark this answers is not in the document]
SamBryce-Smith: Yep line 40-44! That's fair enough, it works perfectly fine so it's not a sticking point for me. 

I'm surprised there isn't a way to make a Nextflow rule 'wait' for all per-sample pre-processing to be done/have a rule that takes all samples as input though. Unless @yuukiiwa or another Nextflow expert wants to chime in XD

execution_workflows/APAtrap/bin/convert_to_tsv.py
```diff
@@ -0,0 +1,44 @@
...
+    df = pd.read_csv(file_in, sep='\t')
+    for index, row in df.iterrows():
+        # write differential file
+        # keep just the gene name
+        name = row['Gene'].split("|")[0]
+        significance = str(row['p.value'])
+        output = [name, significance]
+        differential_out.write("\t".join(output) + "\n")
+
+    differential_out.close()
```
SamBryce-Smith, line 35: When I ran the test data, I get transcript IDs in the first column (e.g. ENSMUSTxxxxx etc.), but I think the [specs want gene IDs](https://github.com/iRNA-COSI/APAeval/blob/main/execution_workflows/execution_output_specification.md#format-03).

I can see in APAtrap's output files that it does not retain the gene ID (ENSMUSGxxx)? I don't know if this will affect summary workflows for the differential usage challenges, but I suspect if we're wanting to match to 'ground truth' this would fail if the ground truth files have gene IDs. I wouldn't be sure how to resolve this either if so, as genes could have different transcript IDs with different p-values.

@ninsch3000 @yuukiiwa @mrgazzara any input on the differential usage ground truth TSVs and whether this is compatible with the summary workflows?

execution_workflows/APAtrap/README.md
```diff
@@ -0,0 +1,80 @@
...
+ - Go to `conf/modules.config` to configure the parameters required to run APAtrap workflow. Descriptions of the parameters
+   are located in the file. Head to `Params` for more info
+ - Delete the existing `results` folder to remove results from the previous run, if any. This is to avoid polluting the next
+   run wtih the results from the previous run. This would likely result in an error as well.
+ - Then, you are good to run the pilot benchmark nextflow pipeline with `APAtrap`
+
+```
+nextflow main.nf --input samplesheet_example_files.csv
+```
+
```
SamBryce-Smith, line 35: Could you add a brief explanation to run the differential usage pipeline too please? i.e. Change config value, make sure to delete results/ if exists and the command (or just say same as above).

execution_workflows/APAtrap/modules/preprocessing.nf
```diff
@@ -0,0 +1,29 @@
...
+
+    output:
+    path(converted_genome_file), emit: ch_genome_file
+
+    script:
+    converted_genome_file = "genemodel.bed"
+
+    """
+    gtfToGenePred $genome_file test.genePhred
+    genePredToBed test.genePhred $converted_genome_file
```
SamBryce-Smith, line 27: Will `test.genePhred` be deleted after the process completes? For tidiness's sake I like to delete intermediate/temporary files but maybe nextflow handles it anyway

execution_workflows/APAtrap/samplesheet_example_files.csv
```diff
@@ -0,0 +1,3 @@
+condition,sample,bam,bai,bed
```
SamBryce-Smith, line 1: Don't need `bed` column in header anymore

execution_workflows/APAtrap/bin/convert_to_bed.py
```diff
@@ -0,0 +1,62 @@
...
+            # same as starting position
+            chromEnd = chromStart
+            score = "."
+            output = [chrom, chromStart, chromEnd, name, score, strand]
+            identification_out.write("\t".join(output) + "\n")
+
+            # write quantification file
+            score = str(row['Group_1_1_Total_Exp'])
+            output = [chrom, chromStart, chromEnd, name, score, strand]
+            quantification_out.write("\t".join(output) + "\n")
```
SamBryce-Smith, line 50: my quantification output looks like this (`control_r1`):
| 16 | 78340249  | 78340249  | ENSMUST00000023572.14 | 85.72  | + |
|----|-----------|-----------|-----------------------|--------|---|
| 16 | 78340414  | 78340414  | ENSMUST00000023572.14 | 85.72  | + |
| 6  | 119924719 | 119924719 | ENSMUST00000088646.11 | 161.84 | - |
| 6  | 119924739 | 119924739 | ENSMUST00000159903.1  | 161.91 | - |
| 6  | 119924260 | 119924260 | ENSMUST00000159903.1  | 161.91 | - |
| 6  | 119924739 | 119924739 | ENSMUST00000177761.7  | 161.91 | - |
| 6  | 119924260 | 119924260 | ENSMUST00000177761.7  | 161.91 | - |

It looks like proximal and distal poly(A) sites of the same txipt are given the same quantification value, I think because you're using the total expression column `Group_1_1_Total_Exp`. I think you may have to use `Group_1_1_Separate_Exp` instead to get per-poly(A) site quantifications.

Saying that, I can also see that the `Group_1_1_Separate_Exp` column always has one fewer poly(A) sites than poly(A) sites reported in `Loci`. e.g. my `deapa` output for `control_r1` in the test data:

| Gene                             | Mean_Squared_Error | Predicted_APA       | Loci                  | Group_1_1_Separate_Exp | Group_1_1_Total_Exp |
|----------------------------------|--------------------|---------------------|-----------------------|------------------------|---------------------|
| ENSMUST00000023572.14\|NA\|16\|+ | 601.5              | 78340249,78340414   | 16:78336188-78340752  | 0.00,35.57,50.15       | 85.72               |
| ENSMUST00000088646.11\|NA\|6\|-  | 2449.3             | 119924719           | 6:119924622-119926687 | 14.76,147.08           | 161.84              |
| ENSMUST00000159903.1\|NA\|6\|-   | 2076.2             | 119924739,119924260 | 6:119923979-119926687 | 4.57,62.48,94.86       | 161.91              |
| ENSMUST00000177761.7\|NA\|6\|-   | 2076.2             | 119924739,119924260 | 6:119923979-119926687 | 4.57,62.48,94.86       | 161.91              |


I think that probably means the 3'end coordinate of the `Loci` column is the most distal poly(A) site and should also added to `apa` and reported in the output files.

Feel free to ask away if any of this is unclear :)

SamBryce-Smith (commented):
Great work yet again @faricazjj! I can run the test runs for both workflows locally after following the instructions in README. I have a few minor comments and some points that I think need wider discussion. To summarise:

Identification & Quantification workflow - approved (with suggested change to script to add most distal site from `Loci` column & per-site quantifications).

Differential usage - would like clarification on whether transcript ID is sufficient or gene ID is required, but that will have to come from others.

When those points are addressed I'm happy to switch to approve. Would also be handy to get a Nextflow expert to have a check at the workflow definition etc. but to the untrained eye I see no problems.

execution_workflows/APAtrap/bin/convert_to_tsv.py
SamBryce-Smith: @faricazjj - We discussed this in the meeting yesterday and decided that if a tool runs a statistical test transcript-by-transcript like here, we will pick the **smallest p-value for the corresponding gene**, and report the gene_id & p-value as according to the [specs](https://github.com/iRNA-COSI/APAeval/blob/main/execution_workflows/execution_output_specification.md#format-03). So I think you will need to do the following:

1. Generate a table mapping transcript_ids to gene_ids from the input GTF file (e.g. like the script in the [DaPars2 workflow here](execution_workflows/DaPars2/Dapars2_snakemake/workflow/scripts/extractIdSymbol.py))
2. Use above table to assign transcript_ids in output files to corresponding gene_id, then pick the smallest p-value for each gene & report those in the differential_usage TSV file.

execution_workflows/APAtrap/samplesheet_example_files.csv
faricazjj: oh yes thank you!

execution_workflows/APAtrap/modules/preprocessing.nf
faricazjj: Good point! Yes test.genePhred is actually not published in the results folder where the other output files that we care about are once the process completes. So it looks like nextflow handles it for us :)

execution_workflows/APAtrap/bin/convert_to_bed.py
faricazjj: @SamBryce-Smith Ah yes good catch! The separate expression column sounds like the right one to get TPM values from. I see, so if I understand correctly, we should get the following mapping from the first row above:

Position       Expression
78340249       0.00
78340414       35.57
78340752       50.15

Is this correct?

I can also contact the author about this later to double check our understanding just to be sure :)

Regarding this, since you mentioned that we should be using separate expression column instead, it just occurred to me that we'll get separate expression column for a different replicate of the same condition. Meaning if we have two replicates, we would get two expression values for each site. Should we then get the average of the two values to put in the final quantification challenge output file? I have also posted the same question on slack in case we need to discuss this further https://apaeval.slack.com/archives/C023SMTS0KS/p1631819510021800

Let me know your thoughts :D

execution_workflows/APAtrap/modules/de_apa.nf
mzavolan: Would make most sense to define the path in the config file, to not run into issues of path resolution rules.

SamBryce-Smith (dismissed):
Looks good to me thank you so much! Some very minor typos/clarifications in the README, once they're done ping me and I'll re-approve. I haven't had a chance to check the identification coordinates in IGV (keeps crashing...), but from what I can see in the code it looks correct (assuming they take the loci from the input BED12 file as is).

dominikburri (dismissed):
Thank you for the great work!
I ran the test data with singularity enabled and both modes worked without problems.
I only have some comments/corrections regarding the readme but they are not urgent.

dominikburri (approved):
Thanks for fixes! Great work :)

dominikburri:
APAtrap is almost done, only one review left. Can maybe @ninsch3000 , @yuukiiwa or @SamBryce-Smith review this PR as well?

SamBryce-Smith:
I can't until Thursday afternoon/Monday but will put it on my schedule for then :)

execution_workflows/APAtrap/README.md
```diff
@@ -0,0 +1,96 @@
...
+51 `singularity.enabled=true` is commented out.
+
+## Singularity
+To run with singularity, comment out line 49 in Apatrap/nextflow.config file `docker.enabled=true` and make sure that line
+51 `singularity.enabled=true` is uncommented.
+
+### Parameters
+Parameters used to run APAtrap are specified in conf/modules.config file. 
+Parameters relevant to the workflow itself are:
+- `run_differentail` - set to false to obtain identification and quantification challenge outputs, or set to true
```
SamBryce-Smith, line 50:
```suggestion
- `run_differential` - set to false to obtain identification and quantification challenge outputs, or set to true
```
Just to fix the typo so it matches exactly with the parameter defined in `conf/modules.config`

SamBryce-Smith (approved):
Looks great thank you again, sorry I didn't get around to this last week.

The one comment is just pedantic me wanting a typo fixed in the README. The only other thing I wanted to raise (I feel like we've discussed this before) is whether APAtrap is compatible with our [new workflow directives/parameters to toggle on/off different challenge files being output by the workflow](https://github.com/iRNA-COSI/APAeval/issues/189). If I remember correctly, APAtrap has to be run differently under the hood in order to perform differential challenge (so it has to be a separate workflow) and is incompatible with the other two challenges. Is it possible to have 'run_identification' and 'run_quantification' as separate true/false parameters to optionally output the files for these challenges?

I don't want the above to block merging but feel it's worth raising as it could save you having to go back and fix this at a later point!

SamBryce-Smith (approved):
Thank you looks good to me! As you said, I think the refactor to have output flags can wait until we've crystallised the new template (and should be a quick fix). Would be good to have another workflow to test out :)

dominikburri (approved):

faricazjj:
@SamBryce-Smith @dominikburri  Thank you so much for reviewing and approving you guys! Merging this now :)

PR #98, opened 2021-06-02
feat: execution workflows APA-Scan & Roar

EuancRNA:
Fixes #26 
Fixes #38 

## Checklist

- [x] I have performed a self-review of my own code
- [x] My code follows the templates/style guidelines of the repository
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have tested my code

PR #161, opened 2021-08-06
feat: Execution workflow for Roar

### LINKED ISSUE #38: Execution Workflow: Roar
## WHAT    
**Write** execution workflow for [Roar](https://github.com/vodkatad/roar/). Use the provided [small files for testing](https://github.com/iRNA-COSI/APAeval/tree/main/tests/test_data) (running the workflow on real data is a different issue).

## CHECKLIST   
* [ ] Comment your code

* [ ] Output: Adhere to output specification for Identification challenge
Fields:

* [ ] Does Roar perform PAS quantification? If yes:

    > - gene ID

SamBryce-Smith:
Fixes #38 

## Checklist

- [ ] I have tested my code

NB: This workflow has already had reviews on the 2 different joint pull requests.
1 - #98
2 - #87

Have moved to separate branches as requested to make it easier to review each method.

Won't copy all comments but these are the main things I (and CJH & AK on the other PR review) I noticed:

- [x] Dockerfile for method (Should be one available here https://quay.io/repository/biocontainers/bioconductor-roar?tab=tags)
- [x] Reformat rule declarations in Snakefile to make more readable
- [x] Ensure test data etc. are deleted from repo
- [ ] Roar.R - Possible 1-off errors by converting Granges to df and outputing BEDs via write.table (GRanges using 1-based indexing, BED does not)
- [ ] Roar requires a custom GTF file, no code to generate but there are instructions in Bioconductor manual. I don't think want to be tied down to specific annotation version and poly(A) site reference, so we should generate as part of pipeline. This would take GTF and reference PolyASite database as input.
- [ ] Roar.R - Org.mm.db is used to find Ensembl IDs which is species specific. Should ideally use input reference GTF to map gene IDs. If generate reference files with workflow (previous point) then this can be ignored.


SamBryce-Smith:
Hey @dominikburri , I've already done a review on the old pull request before I split the branch in two (as did AK & CJH on the PR before that). My main comments are summarised in the bullet points at the top, if you could give it a quick go over and see if I've missed anything. I suspect I'll have to take this one over to address the points

execution_workflows/Roar/workflow/envs/Roar.yaml
```diff
@@ -0,0 +1,65 @@
+name: base
```
dominikburri, line 1: Do we really need this environment  file? Because we would start in the APAeval env anyway, and this includes the ones here (just peeking, did not count). Also, Roar itself is not included, so maybe best to delete this and use the Docker image straightaway.

execution_workflows/Roar/workflow/config/config.Roar.yaml
```diff
@@ -0,0 +1,19 @@
+# Configuration file for snakemake pipeline
```
dominikburri, line 1: This file seems to be duplicated (you mention to delete test data, just want to make sure this is also tagged).

execution_workflows/Roar/workflow/config/samples.csv
```diff
@@ -0,0 +1,3 @@
+sample,sample_type
```
dominikburri, line 1: This file seems to be duplicated (you mention to delete test data, just want to make sure this is also tagged).

execution_workflows/Roar/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+# User manual
+Documentation for Roar is available at:
+
+* Tutorial: https://bioconductor.org/packages/release/bioc/vignettes/roar/inst/doc/roar.pdf
+* User Manual: https://bioconductor.org/packages/release/bioc/manuals/roar/man/roar.pdf
+
+# Environments
+
+**A Docker image still needs to be made for this**. All the needed packages and tools needed to run this are present in the conda `.yaml`, except for Roar itself which can be installed from GitHub/BioConductor (above).
```
dominikburri, line 18: If the docker file you mention (link: https://quay.io/repository/biocontainers/bioconductor-roar?tab=tags) is up-to-date and also supported, then using that Docker image should be sufficient.

execution_workflows/Roar/README.md
```diff
@@ -0,0 +1,47 @@
...
+Config file is `config/config.Roar.yaml`. This contains paths to the `samples.csv` file, output directory, scripts, local & cluster logs and envs (conda `.yaml`'s).
+Samples file present in `config/samples.csv`, and is referred to in `config/config.Roar.yaml`.
+Conda environment, `workflow/envs/Roar.yaml`.
+
+
+
+# Citation
+Please use the following information to cite...
+
+# Contact the Author
```
dominikburri, line 43: I don't think these are the authors of Roar, maybe better to name it: "Contact the Snakemake implementation team" or something similar.

execution_workflows/Roar/README.md
```diff
@@ -0,0 +1,47 @@
...
+# Config files & metadata
+
+Config file is `config/config.Roar.yaml`. This contains paths to the `samples.csv` file, output directory, scripts, local & cluster logs and envs (conda `.yaml`'s).
+Samples file present in `config/samples.csv`, and is referred to in `config/config.Roar.yaml`.
+Conda environment, `workflow/envs/Roar.yaml`.
+
+
+
+# Citation
+Please use the following information to cite...
```
dominikburri, line 41: Could use this:
> Grassi, E., Mariella, E., Lembo, A. et al. Roar: detecting alternative polyadenylation with standard mRNA sequencing libraries. BMC Bioinformatics 17, 423 (2016). https://doi.org/10.1186/s12859-016-1254-8

execution_workflows/Roar/workflow/Snakefile
```diff
@@ -0,0 +1,124 @@
+"""
+Snakemake pipeline for APA-Scan.
```
dominikburri, line 2: Adjust title

execution_workflows/Roar/workflow/Snakefile
```diff
@@ -0,0 +1,124 @@
...
+#-------------------------------------------------------------------------------
+# Parameters, filenames & sample data
+
+#### NOTES #####################################################################
+
+# MAY NEED TO ADJUST THE INDEXING DEPENDING ON THE LEVELS, BUT CAN PROBABLY MAKE
+# THIS CUSTOMISABLE FROM THE COMMAND LINE OR A CONFIG FILE ETC. THIS IS BECAUSE
+# THE LAYOUT OF THESE WILL DICTATE THE ORIENTATION OF LEVELS THAT ARE COMPARED
+# BY ROAR. IE FOR THE TEST DATA, THIS MAKES SO SRSF3 is CONDITION1 AND
+# THE CONTROL IS CONDITION 2
```
dominikburri, line 38: Moving these notes to the README makes it more readable (you mention to reformat rule declarations for better readability, not sure if this is included).

execution_workflows/Roar/workflow/Snakefile
```diff
@@ -0,0 +1,124 @@
...
+################################################################################
+
+
+# Parameters
+THREADS = config["threads"]
+
+
+# Input file patterns
+CONDITION1 = samples["sample_type"].unique()[0]
+CONDITION2 = samples["sample_type"].unique()[1]
```
dominikburri, line 49: Am I correct in assuming that this will use all samples from either `sample_type` condition? That is, if I have 2 samples in condition1 and 3 samples in condition2, all 5 samples are used for the Roar analysis?

execution_workflows/Roar/workflow/scripts/Roar.R
```diff
@@ -0,0 +1,109 @@
...
+### Format output
+
+print(">>>>> SAVING RESULTS")
+# Output df - format 1 (BED) & format 3
+outdf01 <- outdf[,c("seqnames","start","end","source","3utrid","strand")]; outdf01["source"] <- "."
+outdf03 <- outdf[,c("ENSEMBL","pval")]; outdf03 <- outdf03[!duplicated(outdf03),]
+
+# save files
+write.table(outdf01,paste0(out.prefix,"_01.bed"),sep="\t",quote=FALSE,row.names=FALSE,col.names=FALSE)
+write.table(outdf03,paste0(out.prefix,"_03.tsv"),sep="\t",quote=FALSE,row.names=FALSE,col.names=FALSE)
```
dominikburri, line 109: If this part could be separated into another snakemake rule, that  would be great. Then it would be a bit cleaner in that the Roar output is in its conventional form and our specific (and possibly changing) output format is separate from this. This makes it easier to adjust and append on both sides.

dominikburri (commented):
Hi @SamBryce-Smith I went through the files and noted some things not necessarily related to the bullet point comments (especially `Roar.R` I cannot really assess). 

If we could split the main snakemake rule also into pre- and post-processing, this would be great. Not only would it make the code leaner, it also makes it more readable and manageable. However, I do understand that it can be quite some work,  so I don't think it *has* to be done.

execution_workflows/Roar/workflow/scripts/Roar.R
SamBryce-Smith: Great point I completely agree

execution_workflows/Roar/workflow/Snakefile
SamBryce-Smith: I think they are considered as separate groups. The way it is currently set up is to have BAM files for each condition in a separate directory, so this would isolate the two groups. At [lines 15 & 16 in the Roar.R script](https://github.com/iRNA-COSI/APAeval/blob/da00a5b31e730a54b2f0c79adfddfb0cec83c902/execution_workflows/Roar/workflow/scripts/Roar.R#L15) BAM files for each condition are defined separately (saying that I don't like how it's done (they seem to be combined again?), will comment on the actual script).

Roar can only compare between two conditions (I'm almost certain but will double check) so we should also include a check that only two conditions are specified in the sample sheet

execution_workflows/Roar/workflow/scripts/Roar.R
```diff
@@ -0,0 +1,109 @@
...
+
+# Input gtf
+gtf <- args[1]
+
+# Input bams
+bam1 <- paste(path=args[2],list.files(path=args[2],pattern="*.bam"),sep="/")
+bam2 <- paste(path=args[3],list.files(path=args[3],pattern="*.bam"),sep="/")
+bams <- c(bam1, bam2)
+# Output prefix
+out.prefix <- args[4]
```
SamBryce-Smith, line 19: So BAM files are collected separately for each condition, then combined into a single vector? I think they should be kept separate as files for each condition are separate arguments for the Roar input function

I'd prefer to have an 'argparse' style input/CLI though, where files are passed for each condition as a separate option (e.g. --condition1 a.bam b.bam --condition2 c.bam d.bam)

execution_workflows/Roar/workflow/scripts/Roar.R
```diff
@@ -0,0 +1,109 @@
...
+print(paste0("   >> BAM2: ",bam2))
+
+
+
+
+### Create RoarData Object
+
+# Generate RoarDataSet
+print(">>>>> GENERATING ROAR DATA SET")
+rds <- RoarDatasetMultipleAPAFromFiles(bams[1:2], bams[3:4], gtf)
```
SamBryce-Smith, line 39: As with above this should be changed. This hardcodes each condition to have two samples only without checking which condition they arise from, so is only valid when have a 2*2 comparison.

We should also consider having the option to specify what is the 'base' condition. I don't think this is very clear from the sample table/README

SamBryce-Smith:
Splitting up required tasks broadly into 'workflow-related', 'code-related' and 'documentation-related', so people with no Snakemake experience can still help out with the workflow! NB the list may not be completely exhaustive, but there should be enoguh

Code related:

- [ ] Script to generate Roar-compliant GTF file from a reference GTF file and a database of poly(A) sites. 

     - There are instructions available on [Bioconductor vignette (2.1 Alternative polyadenylation annotations)](https://bioconductor.org/packages/release/bioc/vignettes/roar/inst/doc/roar.pdf)
  
- [ ] Roar.R script should just contain steps to run Roar. Create second script to parse output to APAeval metric/output file formats
     
     - Can take code from second half of current Roar.R script, but be careful about GRanges --> write.table() to create BED file - this could introduce a 1-off error with coordinates as GRanges coordinates are 1-based,end inclusive (recommend using rtracklayer's export instead).

- [ ] Address comments by @SamBryce-Smith in `Roar.R` RE how BAM files per condition are passed to Roar

    - Ideally, change CLI so pass desired BAM files as separate options (e.g. --condition1 a.bam b.bam --condition2 c.bam d.bam)
    - Keep BAM files per condition as separate objects, remove limit to two samples per condition

- [ ] Dockerfile for Roar dependencies

    - This may be most straightforward https://quay.io/repository/biocontainers/bioconductor-roar?tab=tags

Docs related:

- [ ]  Delete test files from repository (use test data in main repo instead)
- [ ] Add citation to bottom of README

SamBryce-Smith:
I think this is ready for review now! Test run works successfully on my laptop using the test data in the main repository.

image

A few notes:
- (I think) incompatible with both identification (uses/takes all provided reference sites, no de-novo prediction) and quantification (no TPM quantification provided). 
- I've used biocontainers for all dependencies so nothing extra to be uploaded to team dockerhub

**Generating Roar-compliant annotations*

I've written a custom script to generate the custom GTF annotation file (`workflow/scripts/make_roar_annotations.py`) since no scripts/functions are provided to the best of my knowledge. I've tried to follow the instructions as best as I can, but 2nd/3rd/4th opinions on this would be appreciated :)). I haven't been able to fully validate my workflow vs their provided annotations as I cannot work out how to specifically download RefSeq v79 transcript annotations (maybe I'm being blind but I was trawling through the ftp site and had no luck). 

I've only written code to generate 'single APA type' annotations so far, although ['multiple APA'](https://github.com/vodkatad/roar/wiki/Identify-differential-APA-usage-from-RNA-seq-alignments) annotations (multiple polyA sites per gene) can be provided. Initially I though the instructions ([section 3.6, pages 6-7](https://bioconductor.org/packages/release/bioc/vignettes/roar/inst/doc/roar.pdf)) were tricky to follow, but after a re-read and some browsing of the provided file I think this shouldn't be too tricky (if anything it's probably the easier one to generate!). If we consider this important I'll happily implement it (will also have to update the `Roar.R` script but shouldn't take too much time).

**Selecting p-value to report**

I somewhat lazily chose the 'pval' column, which according to the [docs (section 3.5)](https://bioconductor.org/packages/release/bioc/vignettes/roar/inst/doc/roar.pdf) represents:
>The pval column stores the results of the Fisher tests when
both conditions have a single sample, while the multiplication of all the possible
combinations p-values for un-paired multiple samples analysis or the combined
p-values following the Fisher method if computePairedPvals has been used.

Perhaps that is fine, but have to admit I've not come across p-value multiplication in a tool before... I think Roar performs a 1v1 Fishers test on every possible combo of control v treatment samples, so another option would be to pick the smallest p-value across comparisons to report, but I don't think that is any more/less valid...

Feedback much appreciated :)

dominikburri (commented):
Hi @SamBryce-Smith , thanks so much for the work!
Can you also tick the respective boxes in the [original description](https://github.com/iRNA-COSI/APAeval/pull/161#issue-962748407) of this PR?

I couldn't go through all the code for now, but wanted to give early feedback. 
You mention validating the workflow, maybe it's good if you test the workflow with the annotations we are using in the benchmark on samples you know how the results look like. But then again, it might be unfair.. difficult.

## regarding p-values
multiplication of p-values doesn't really make sense in my opinion. With the test data as you supplied it, I get a value < `1e-200` for either gene. But when I remove the duplicate samples and compare only one sample per condition, the value is up to ca. `1e-53`. Maybe they intended to have a geometric mean? Imagine having 5 samples per condition, each with a p-value of 0.5, will give you a significant hit: `0.5^5 = 0.03`.
Not sure how to deal with this, maybe it's best what you suggested and just take the minimum or maximum.

## regarding gtf generation
I followed the notes and quotes in `make_roar_annotations.py`. 
-  The "single" APA just means that a proximal and distal site will be extracted, correct?
- *multiAPA* means that there are `>= 3` polyA sites (PAS), correct? Why would it be easier to generate? I mean it would involve finding more PAS from the annotation and possibly more tests, as now multiple proximal or distal sites are there. 
- The gtf with polyA sites is always stranded. When stranded parameter is used, only reads with correct orientation are counted? And when unstranded, all reads intersecting the polyA site (irrespective of orientation) are used?

execution_workflows/Roar/workflow/scripts/make_roar_annotations.py
```diff
@@ -0,0 +1,563 @@
...
+
+"Every gene with at least one APA site is considered; their
+longest transcript is used to get the “classical” polyadenylation syte (canonical
+end of the longest transcript stored in UCSC) and the most proximal (with
+respect to the TSS, that is the farther to the canonical end of the transcript;
+but preferring sites still in the 3’UTR) reported APA site found in PolyA DB is
+used to define the end of the short isoform"
+
+1. Overlap GTF with polyA sites, retain gene IDs that have at least 1 overlapping polyA site
+2. Select the longest transcript for each gene
```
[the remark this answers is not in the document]
SamBryce-Smith: I completely agree it doesn't make the most sense. I tried to interpret the [documentation](https://www.bioconductor.org/packages/release/bioc/vignettes/roar/inst/doc/roar.pdf) as literally as possible which says the following:

> Every gene with at least one APA site is considered; **their
longest transcript** is used to get the “classical” polyadenylation syte (canonical
end of the longest transcript stored in UCSC) and the most proximal (with
respect to the TSS, that is the farther to the canonical end of the transcript;
but preferring sites still in the 3’UTR) reported APA site found in PolyA DB is
used to define the end of the short isoform

Perhaps it's just an omission and it is meant to be the longest 3'UTR (or the most 'downstream' 3'UTR). I can't see a way to interpret what they've described here differently to how I've implemented it, but then again I've spent a lot of time looking at this so I could well be acting hardheadedly. 

I could get in touch with the author to clarify but think it's unlikely we'll get a response within our timeline (worth a try regardless I suspect). I think the only definitive way to answer this would be if I could regenerate the [provided annotations](https://github.com/vodkatad/roar/wiki/Identify-differential-APA-usage-from-RNA-seq-alignments#hg38-annotation-files-refseq-version-79) and do some comparisons, but I previously had a hard time finding RefSeq v79 GTF...

Thoughts on what I should do next @dominikburri @faricazjj ?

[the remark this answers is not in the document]
dominikburri: Thanks for putting the documentation snippet here :) 
I'm in clinch: should we take it literal and use the longest transcript or interpret it as the longest 3'UTR, just as you describe it.

So how about we keep it as is for now and we contact them? If we hear from them we can still make changes afterwards. I mean if it involves re-running some results, this should be doable (I hope so at least). 
You can also ask if they can give you RefSeq v79 GTF to actually regenerate the annotations you mention. 

What we can also try to do is to simply check in IGV how the resulting GTF looks like for the human or mouse genome and whether the polyA sites roughly overlap with polyA site databases or annotated 3'UTR features... Though not exactly sure what this would gain us.

SamBryce-Smith:
image

I've rejigged a bit to reflect fact we are going to use pre-provided annotations, not all comments addressed yet. Few new problems arising now. Firstly, pre-provided mouse annotations are mm9 & our test data are mm10 so results with our test data are nonsense (all p-values are 1). Secondly, Roar's custom annotations use RefSeq gene names. For outputs to be compatible with execution workflow specifications I will need to convert to Ensembl gene IDs, possibly using something like Biomart unless anyone else has any suggestions. I haven't addressed @dominikburri's comments regarding which p-values to report yet.

PR #177, opened 2021-09-22
Add test files with 'chr' prefix added to seqnames

SamBryce-Smith:
Fixes #{ISSUE_NUMBER}

## Checklist

- [x] I have tested my code

Adds a set of test files with the `chr` prefix added to chromosome names. This is required to [test DaPars](https://github.com/iRNA-COSI/APAeval/pull/169#discussion_r697756991) which hardcodes for the `chr` prefix. As we intend to use Gencode versions of transcript annotations and genome assemblies, this also adds test data that matches the chromosome naming convention we will use in the challenge.

I've added the `chr` prefix to the GFF3, GTF, BED and BAM files. The GFF3, GTF & BED files were modified with the following awk command:

```
awk '{if($1 !~ /^#/) {print "chr"$0}}  else {print $0}' input > output 
```

For BAM files, the chr prefix was added using the following commands. After modification, the BAM files were re-sorted and indexed:
```
samtools view -h input.bam | \
sed -e '/^@SQ/s/SN\:/SN\:chr/' -e '/^[^@]/s/\t/\tchr/2' | \
samtools view -bS - | samtools sort - > out.bam && samtools index out.bam
```

I was able to successfully run the prefixed test files (BAMs & GTF file) through the DaPars2 execution workflow without error (currently on the `DaPars2_snakemake` branch). I have not yet tested the GFF3 file or BED file as I do not know of a workflow that uses these files.
image

@faricazjj - it may be worth you testing these files with your DaPars workflow as it currently is?


faricazjj (approved):
LGTM thanks so much! :)

ninsch3000 (approved):

Issue #19, opened 2021-05-24
Execution Workflow: APAlyzer


ninsch3000:
## WHAT    
**Write** execution workflow for [APAlyzer](https://bioconductor.org/packages/release/bioc/html/APAlyzer.html). Use the provided [small files for testing](https://github.com/iRNA-COSI/APAeval/tree/main/tests/test_data) (running the workflow on real data is a different issue).

## CHECKLIST   
* [x] Use [snakemake template](https://github.com/iRNA-COSI/APAeval/tree/main/docs/templates/snakemake) or [nextflow template](https://github.com/iRNA-COSI/APAeval/tree/main/docs/templates/nextflow) to create your workflow.
* [x] Comment your code
* [x] Run individual rules/processes in either conda envs or docker/singularity containers for reproducibility
* [x] Input:  .bam or .fastq from [`test_data`](https://github.com/iRNA-COSI/APAeval/tree/main/tests/test_data)
* [ ] Give [feedback about the method](https://forms.gle/mAAkbRJqhedJnSeJ6)

## OUTPUTS (see [specification](https://github.com/iRNA-COSI/APAeval/blob/main/execution_workflows/execution_output_specification.md)): 
* [x] ~Output: Adhere to output specification for Identification challenge~ (No identification)
* [x] ~Output: Adhere to output specification for quantification challenge~ (No quantification)
* [x] Output: Adhere to output specification for differential usage challenge
   > This TSV file contains two columns:

    > - gene ID
    > - significance of differential PAS usage

    > Column names should not be added to the file.

mkatsanto:
I have created the following branch https://github.com/iRNA-COSI/APAeval/tree/apalyzer
APAlyzer implementation is currently stuck since the creation of a reference genome based on a given gtf is not working for this tool. You currently have to rely on predefined PAS atlas, which is not reliable for the benchmark.
I opened an issue in the git repo of the tool (https://github.com/RJWANGbioinfo/APAlyzer/issues/7) and await further info.

SamBryce-Smith:
@mkatsanto looks like [issue you've reported](https://github.com/RJWANGbioinfo/APAlyzer/issues/7) has been fixed. Are you still able to carry on with completing this workflow (looks like you've done a lot already!) or should we consider this workflow open for other contributors?

Thanks :)

SamBryce-Smith:
[...]

PR #180, opened 2021-09-29
Apalyzer

mkatsanto:
Fixes partially  #19 

There are still some issues that block the merge of apalyzer. 
The first issue is the one I raised here:  https://github.com/RJWANGbioinfo/APAlyzer/issues/7   which was fixed by the author.

- The update of the code that was made by the apalyzer author is not present in a conda env (since this is part of bioconductor so the updates do not happen so often) so we can only have the docker implementation. The dockerfile.yaml contains the instructions on how to make this.
- In this code you still need to provide genome annotation from ensembl and the name of the gtf file should have the same format as the ensembl name, otherwise this does not work. In the config I refer to a mouse gtf file from ensembl but did not add this in the repo so as not to burden the git. 
- When I tried with a whole gtf the code was running, however when I made a small gtf file out of the ensembl file this fails. Perhaps we should open another issue in the authors' git.


faricazjj:
Closing this PR, the complete APAlyzer execution workflow is added in this [PR](https://github.com/iRNA-COSI/APAeval/pull/195). Thanks @mkatsanto ! :)

PR #182, opened 2021-10-04
update quantification summary workflow content

yuukiiwa:
I have updated the content of the quantification summary workflow with the Q2 quantification metrics and other fixes made by Asier and me. 

To add in other metrics for the quantification challenge, one can add in a script with python functions that can be called from [`benchmarking_dockers/apaeval_metrics/compute_metrics.py`](https://github.com/iRNA-COSI/APAeval/blob/5e4c577371cc6f4f032ad4d1b1ccc26ff589556a/summary_workflows/quantification/benchmarking_dockers/apaeval_metrics/compute_metrics.py#L11). Getting [the json assessment files](https://github.com/iRNA-COSI/APAeval/blob/5e4c577371cc6f4f032ad4d1b1ccc26ff589556a/summary_workflows/quantification/benchmarking_dockers/apaeval_metrics/compute_metrics.py#L69) may need a little more thoughts on as they determine how the plots look like on OpenEBench.

Thank you!!


summary_workflows/quantification/benchmarking_dockers/README.md
```diff
@@ -1,6 +1,6 @@
-# TCGA benchmarking Docker declarations
+# APAeval benchmarking Docker declarations
 
-OpenEBench TCGA benchmarking Docker declarations, which define the architecture of benchmarking workflows to be implemented in OpenEBench.
+OpenEBench/iRNA COSI APAeval benchmarking Docker declarations, which define the architecture of benchmarking workflows to be implemented in OpenEBench.
 
 **NOTE for developers.** In order to make the workflow containers reproducible and stable in the long-term, make sure to use specific versions in the container base image (e.g.*ubuntu:16.04*, NOT *ubuntu:latest*).
```
ninsch3000, line 6: In the following section, links have to be adapted to apaeval (are still tcga) (I can't directly comment within the section for whatever reason, but I see 5 occurrences of tcga)

summary_workflows/quantification/benchmarking_dockers/README.md
```diff
@@ -20,14 +20,7 @@ the benchmark itself is performed by merging the assessment metrics with the res
 SVG format - scatter plot, and JSON format - aggregation/summary datasets, which are also compatible with the [Elixir
```
ninsch3000, line 20: This sentence isn't too clear

summary_workflows/quantification/benchmarking_dockers/apaeval_validation/validation.py
```diff
@@ -2,6 +2,7 @@
 
 from __future__ import division, print_function
 import pandas
+import numpy as np
 import os, json
 import sys
 from argparse import ArgumentParser
```
ninsch3000, line 8: The argument declaration below still contains descriptions for cancer data instead of APAeval

ninsch3000 (changes requested):
Thanks a lot @yuukiiwa!

I tried to review, but I'm afraid I still don't understand everything. However, here are some comments (Please ignore if they don't make sense!):
- Could the directory `Q2_benchmark` be renamed to `Q2_specification`? As I understand, we don't need it for anything else than the specification, correct? (This renaming would also have to be done for the metrics in the other two events). The `original_Q2_scripts` could probably also be removed, as they are now merged in `Leo_MatchPAsites_GT.py`, which also nicely contains the corresponding README.
- just for clarity, can the directory `benchmarking_dockers` be renamed to `quantification_dockers`? Do I understand correctly, that each additional metric (Q3, Q4, etc) would be added as python scripts inside the `apaeval_metrics`, similarly to the `Leo_MatchPAsites_GT.py`? That in turn means, that for *all* quantification challenges, *all* the metrics will be calculated, right? So just in case we had one metric Qx, which could only be applied to a certain dataset (e.g. the simulated data), we'd have to create an extra summary workflow for Qx? If that is the case, renaming of `benchmarking_dockers` to `quantification_dockers_Q2` might make more sense, and if we add other metrics, we update the name to `quantification_dockers_Q23`, and we have the special `quantification_dockers_Qx` separately. (Sorry if this doesn't make sense)
- There are still some places, where TCGA/cancer data is mentioned, I tried to mark them, but probably didn't catch all.
- Could we have a README for the `APAeval_Q2_test_data` directory, explaining where the data came from (and then possibly renaming the files themselves to test_input1, test_input2, etc.). In particular, it might be useful to explain where these data have to reside for a real OEB run (same directory structure needed?), or whether there are any naming requirements, or in which config they have to be provided, or similar. Honestly I currently don't really understand the difference between the bed files in this dir and those in the child dir `metrics_ref_datasets` and why there is no bed in data, and so on.
- the `build.sh` isn't gonna work, as it contains tcga names (Couldn't directly comment on it because it is unchanged)

So that's just some superficial things I noticed now, I haven't done a testrun yet.
Thanks  Yuk Kei <3

summary_workflows/quantification/main.nf
```diff
@@ -3,23 +3,26 @@
...
 	    Usage:
 	    Run the pipeline with default parameters:
 	    nextflow run main.nf -profile docker
 	    Run with user parameters:
- 	    nextflow run main.nf -profile docker --input {driver.genes.file} --public_ref_dir {validation.reference.file} --participant_id {tool.name} --goldstandard_dir {gold.standards.dir} --cancer_types {analyzed.cancer.types} --assess_dir {benchmark.data.dir} --results_dir {output.dir}
+ 	    nextflow run main.nf -profile docker --input {driver.genes.file} --public_ref_dir {validation.reference.file} --participant_id {tool.name} --goldstandard_dir {gold.standards.dir} --challenges_ids {analyzed.challenges} --assess_dir {benchmark.data.dir} --results_dir {output.dir}
 	    Mandatory arguments:
                 --input		List of cancer genes prediction
 				--community_id			Name or OEB permanent ID for the benchmarking community
                 --public_ref_dir 		Directory with list of cancer genes used to validate the predictions
```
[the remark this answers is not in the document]
AsierGonzalez: Well, what I meant with my comment was simply that the description `--public_ref_dir 		Directory with list of cancer genes used to validate the predictions` does not apply to what the APAeval data is about. How the validation is done is a topic for a later comment. If we don't validate against an external reference file, which this is supposed to do, we may decide to remove this parameter altogether.

summary_workflows/quantification/main.nf
```diff
@@ -3,23 +3,26 @@
...
                 --participant_id  		Name of the tool used for prediction
                 --goldstandard_dir 		Dir that contains metrics reference datasets for all cancer types
```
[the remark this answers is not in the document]
AsierGonzalez: Thank you! Unfortunately, now the phrase `Dir that contains (...)` is repeated [(line 24)](https://github.com/iRNA-COSI/APAeval/blob/add_working_Q2_summaryworkflow/summary_workflows/quantification/main.nf#L24):

```
--goldstandard_dir 		Dir that contains Dir that contains gold standard/ground truth files used to calculate the metrics for all challenges
```

summary_workflows/quantification/quantification_dockers/apaeval_consolidation/manage_assessment_data.py
```diff
@@ -9,7 +9,7 @@
 import numpy as np
 from assessment_chart import assessment_chart
 
-DEFAULT_eventMark = '2018-04-05'
+DEFAULT_eventMark = '2021-06-04'
```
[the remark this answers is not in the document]
AsierGonzalez: Uuuum, I was thinking about defining it in `nextflow.config` and passing it as a parameter to `manage_assessment_data.py` in [main.nf](https://github.com/iRNA-COSI/APAeval/blob/add_working_Q2_summaryworkflow/summary_workflows/quantification/main.nf#L151) rather than creating a new python function that we need to update separately. Do you think this is a better option?

summary_workflows/quantification/README.md
```diff
@@ -1,41 +1,35 @@
...
-        13. weight of prediction expression
-
-### Usage:
-
-```bash
-python3 match_with_gt.py -o output.bed prediction.bed ground_truth.bed window_size
-
-# example:
-python3 match_with_gt.py adultCortex.PAPERCLIP.mm10.bed siControl_R1.MACEseq.mm10.bed 15
+# OpenEBench assessment workflow in Nextflow
```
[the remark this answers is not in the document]
AsierGonzalez: So now both [quantification/README.md](https://github.com/iRNA-COSI/APAeval/blob/add_working_Q2_summaryworkflow/summary_workflows/quantification/README.md) and [quantification_dockers/apaeval_metrics/Leo_MatchPAsites_GT/README.md](https://github.com/iRNA-COSI/APAeval/blob/add_working_Q2_summaryworkflow/summary_workflows/quantification/quantification_dockers/apaeval_metrics/Leo_MatchPAsites_GT/README.md) contain basically the same information? I liked that in the [previous version](https://github.com/iRNA-COSI/APAeval/blob/9bdd0b658cc933e65f16855c333da67ccdf1bec3/summary_workflows/quantification/README.md) `quantification/README.md` described the three steps in the summary workflow and also how the docker containers could be built locally. I think we should recover this content here, keep the new part `Running the summary workflow` and remove `Matching PA sites to ground truth`.

summary_workflows/quantification/quantification_dockers/apaeval_consolidation/manage_assessment_data.py
```diff
@@ -8,8 +8,9 @@
 from argparse import ArgumentParser
 import numpy as np
 from assessment_chart import assessment_chart
+from assessment_chart import conf
 
-DEFAULT_eventMark = '2018-04-05'
+DEFAULT_eventMark = DEFAULT_eventMark()
```
AsierGonzalez, line 13: This assignment makes me feel uneasy as it feels very error-prone to name the variable exactly as the function. However, if you say you did it on purpose I can live with it.

AsierGonzalez:
> Hi @AsierGonzalez, thank you so much for the very detailed review!! :) I have made changes based on your suggestions. The only thing that I didn't add in is the `docker_local` profile because nextflow recognizes the tag of a local container and does not pull one if a container exists locally.

Thank you @yuukiiwa  for addressing my comments duly, great work! I still think that there are a few of them that have not been resolved completely, see my responses.

Regarding the need or not to have a `docker_local` profile, I get your point but I don't think that it works at the moment since the names/tags of the local and remote containers are different:
- Local containers: `apaeval_validation`, `apaeval_metrics`, `apaeval_consolidation` as defined in [build.sh](https://github.com/iRNA-COSI/APAeval/blob/add_working_Q2_summaryworkflow/summary_workflows/quantification/quantification_dockers/build.sh#L13).
- Remote containers: To start with all the container names downloaded from Docker Hub belong to the `apaeval` namespace and, in addition, the names are also different: `apaeval/q2_validation`, `apaeval/q2_metrics`, `apaeval/q2_consolidation`.

OrAa I missing something and I am wrong?

summary_workflows/quantification/README.md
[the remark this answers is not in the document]
yuukiiwa: I have reverted the `quantification/README.md` content and added `Running the summary workflow` section to the `quantification/README.md`

summary_workflows/quantification/main.nf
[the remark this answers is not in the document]
yuukiiwa: Thanks for the clarification! I have changed the `--public_ref_dir` definition

summary_workflows/quantification/main.nf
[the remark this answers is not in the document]
yuukiiwa: Thanks for catching this! I have removed the repeated part of the sentence

summary_workflows/quantification/quantification_dockers/apaeval_consolidation/manage_assessment_data.py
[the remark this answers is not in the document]
yuukiiwa: That's a good idea! The `nextflow.config` now has a parameter called `eventMark`, which is passed to `manage_assessment_data.py` through an argument. The `docker` container is also updated and tested.

summary_workflows/quantification/quantification_dockers/apaeval_consolidation/manage_assessment_data.py
yuukiiwa: I find this weird too, so I just have it removed.

yuukiiwa:
Thank you for the suggestions, @AsierGonzalez! I have addressed them and ran the summary workflow locally with the updated q2_consolidation `docker` container:
Screenshot from 2021-11-05 11-16-15

Regarding the local docker naming, do you think it is a good idea to change the docker directory names from `apaeval_<step>` to `<benchmark (ex. q2)>_<step>` so that we can keep the names of the local `docker` containers consistent with those of the remote `docker` containers?

summary_workflows/quantification/quantification_dockers/apaeval_consolidation/manage_assessment_data.py
AsierGonzalez: Thank you! And well done removing the definition from the global scope, I hadn't realised it!

One last thing about this, do you think that `DEFAULT_eventMark` is a good name for a variable that currently holds a date? Honestly, I don't know what "eventMark" is supposed to mean, perhaps "event_date" would be more meaningful and the "DEFAULT" prefix can be removed definitely.

yuukiiwa: Thanks for the suggestion! I have changed `DEFAULT_eventMark` to `event_date`. Also, the docker directories are now named `q2_<validation/consolidation/metrics>` instead of `apaeval_<validation/consolidation/metrics>`.

summary_workflows/quantification/quantification_dockers/q2_consolidation/assessment_chart/assessment_chart.py
```diff
@@ -242,12 +242,12 @@ def print_chart(cancer_dir, summary_dir, cancer_type, classification_type):
...
 
-    plt.title("Cancer Driver Genes prediction benchmarking - " + cancer_type, fontsize=18, fontweight='bold')
+    plt.title("Quantification benchmark 2 - number of matched sites vs number of unmatched sites in challenge" + challenge_type, fontsize=18, fontweight='bold')
 
     # set plot title depending on the analysed tool
 
-    ax.set_xlabel("True Positive Rate - % driver genes correctly predicted", fontsize=12)
-    ax.set_ylabel("Precision - % true positives over total predicted", fontsize=12)
+    ax.set_xlabel("Number of matched sites", fontsize=12)
+    ax.set_ylabel("Number of unmatched sites", fontsize=12)
```
daneckaw, line 250: Numbers of matched and unmatched sites are not relevant for quantification, but may be used in identification benchmarks. The only metric in this benchmark should be correlation.
However, if we want to use it in Q2 for now anyway, it would me more informative to calculate the percentage of matched sites, e.g. %matched = matched / (matched + unmatched) and display it as a barplot.

summary_workflows/quantification/quantification_dockers/README.md
```diff
@@ -0,0 +1,24 @@
...
+performance metrics - precision (Positive Predictive Value) and recall(True Positive Rate). Those metrics are written
+into assessment datasets. In order to create datasets with structure compatible with the [Elixir
+    Benchmarking Data Model](https://github.com/inab/benchmarking-data-model), please use the following [python module and JSON schema](./apaeval_metrics/JSON_templates)
+3. [**Consolidation**](./apaeval_consolidation):
+the benchmark itself is performed by merging the assessment metrics with the rest of the predictions generated from APAeval execution workflows. The results are provided in both SVG format (scatter plot) and JSON format (aggregation/summary datasets that are compatible with the [Elixir Benchmarking Data Model](https://github.com/inab/benchmarking-data-model))
+
+For the tool execution workflows please refer to the [iRNA-COSI/APAeval/execution_workflows](https://github.com/iRNA-COSI/APAeval/tree/main/execution_workflows) repository.
+
+## Usage
+In order to build the Docker images locally, please run `./build.sh <tag_id>`
```
dominikburri, line 24: in `nextflow.config`, the tag is hardcoded for **1.0**. Will this cause troubles?

summary_workflows/quantification/quantification_dockers/q2_metrics/compute_metrics.py
```diff
@@ -54,7 +54,7 @@ def compute_metrics(input_participant, gold_standards_dir, challenge_types, part
         # get metrics dataset
 #        metrics_data = pandas.read_csv(os.path.join(gold_standards_dir, challenge + ".txt"),
 #                                       comment="#", header=0, sep='\t',index_col=0)
-        metrics_data=os.path.join(gold_standards_dir, "siControl_R2.MACEseq.mm10.bed") ##PLEASE CHANGE THIS LATER
+        metrics_data=os.path.join(gold_standards_dir, challenge + ".bed") ##NO LONGER HARD-CODED
 
         # metric on the number of matched sites
         window = 15
```
dominikburri, line 60: Why is the window size hardcoded? Do we want to change that?

dominikburri (commented):
Hi @yuukiiwa, @AsierGonzalez and @ninsch3000 , thank you very much for the hard work!
I tried to run the test but I couldn't get my Mac to run the Docker images. Maybe I'm missing something, but the goal is to run it on Linux anyway. 

Nevertheless, I have some comments and questions.

* Does each challenge (here **Q2**) reside in its own subdirectories and three steps (*validation*, *metrics*, *consolidation*)? That is, for each challenge three new Docker images will be generated? I know this is basically the [same question as before](https://github.com/iRNA-COSI/APAeval/pull/182#issuecomment-944086222), but I want to be clear about it.
  * e.g. in line 52 in `nextflow.config` there is a parameter *challenge_ids* which sounds like could take several challenges at once. Is this intended to be extended as list?
*  I agree with @AsierGonzalez that a new profile with local containers should be created. see [this comment](https://github.com/iRNA-COSI/APAeval/pull/182#pullrequestreview-784681529). Right now local containers are not considered in the workflow.
* As @daneckaw [pointed out](https://github.com/iRNA-COSI/APAeval/pull/182#discussion_r756992289), I also think there is a mismatch in what `Q2_benchmark_specification.md` specifies as metric and what is computed and outputted in `compute_metrics.py`. The former defines solely correlation but the latter 2 additional metrics. Unfortunately I cannot test it right now, but will come back to it.

summary_workflows/quantification/main.nf
```diff
@@ -3,23 +3,26 @@
...
+	    APAeval QUANTIFICATION BENCHMARKING PIPELINE
+		Author(s): Yuk Kei Wan (*1,*2), Asier Gonzalez (*3)
+		*1 iRNA COSI
+		*2 Genomic Institute of Singapore, A*STAR, Singapore
+		*3 Barcelona Supercomputing Center, Barcelone, Spain
+		2021
+	    =====================================================
 	    Usage:
 	    Run the pipeline with default parameters:
 	    nextflow run main.nf -profile docker
```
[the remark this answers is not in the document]
AsierGonzalez: > Took me a while to find the values of the default parameters, they are in nextflow.config. Maybe adding a pointer for oblivious snakemake users would be helpful :)

Do you think it would be enough if we changed row 15 to something like this?:
```
Run the pipeline with default parameters read from nextflow.config :
```
If so we may also want to update the last sentence in the [README](https://github.com/iRNA-COSI/APAeval/blob/add_working_Q2_summaryworkflow/summary_workflows/quantification/README.md#running-the-summary-workflow) accordingly:
```
One can use the following command to run the summary workflow from command line:

nextflow run main.nf -profile docker
```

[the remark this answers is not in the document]
AsierGonzalez: > And am I correct in assuming that the default parameter values are left untouched, but the nextflow call would be modified with the mandatory arguments? That is, a valid run on real data must always have the mandatory arguments below?

I'm not quite sure I understand what you mean here. Both runs with test and real data have to specify all the arguments. The difference is that with test data they are all specified in `nextflow.config` whereas when we run it in the OEB environment we overwrite some of the values on the command-line, which takes precedence. This is explained in the [Nextflow docs](https://www.nextflow.io/docs/latest/config.html#configuration-file).

summary_workflows/quantification/quantification_dockers/README.md
AsierGonzalez: You are right, it does. At the moment the docker images **must** be tagged as `1.0`. When we were making changes @yuukiiwa would tag all the images with that tag but this is not ideal. We should either update the README to remind the user that the tag there must match the one in `nextflow.config` or add some logic somewhere to sync both.

summary_workflows/quantification/Q2_specification/Q2_benchmark_specification.md
```diff
@@ -0,0 +1,90 @@
+# Q2 benchmark specification
```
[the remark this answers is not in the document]
AsierGonzalez: Good question - you are right in that they seem to be identical. It seems that `Q2_benchmark` was there before and `Q2_specification` appeared later on. I vaguely remember that at the time @daneckaw made some changes to [this same directory in master](https://github.com/iRNA-COSI/APAeval/tree/main/summary_workflows/quantification/Q2_benchmark), perhaps there was some synchronisation merge that duplicated the content.
Anyhow, we should decide which one we want to keep, one should go.

summary_workflows/quantification/README.md
```diff
@@ -1,41 +1,42 @@
...
+    - Inputs the `output.json` file from the metrics computation step and the `data/` directory, which stores files with benchmark values
+    - The current python scripts are as they are in https://github.com/inab/TCGA_benchmarking_dockers, and only supports 2D plots with x and y axes
+
+#### After making the necessary changes for your specific challenge, you will have to build the docker image locally
+Go to the specific docker directory for each step in `benchmarking_dockers/`:
+ - `apaeval_validation/`, `apaeval_metrics/`, and `apaeval_consolidation/`
+and type the following
 ```
+docker build . -t apaeval_[challenge]_[validation/metrics/consolidation]:1.0
+```
```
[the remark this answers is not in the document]
AsierGonzalez: I agree that it isn't consistent to mention here that images can be built using `docker build` and then have the [README in quantification_dockers](https://github.com/iRNA-COSI/APAeval/blob/add_working_Q2_summaryworkflow/summary_workflows/quantification/quantification_dockers/README.md) the script `build.sh` to do the same thing. Probably both approaches should be mentioned in only one README.

AsierGonzalez:
Thank you for your comments @dominikburri! I have added responses to your comments where I could add something but some of them need input from @yuukiiwa and @daneckaw.

Regarding your other questions in your review comment here is what I can say about two of them:

>   Does each challenge (here Q2) reside in its own subdirectories and three steps (validation, metrics, consolidation)? That is, for each challenge three new Docker images will be generated?

From the discussion with @ninsch3000 I think that this will be the case. I believe we agreed that identification, quantification and differential usage will be different "benchmarking events" rather than "challenges", but what you say is true. It would be possible to reuse some of the containers if the same code would work for two different benchmarking events but I think the input files, metrics and plots will be different enough to justify keeping the code separate. Note that this doesn't mean that the code will be used only once because each benchmarking event will have multiple challenges (i.e. reference files) and we should be able to use exactly the same summary workflow for all of them.

> e.g. in line 52 in nextflow.config there is a parameter challenge_ids which sounds like could take several challenges at once. Is this intended to be extended as list?

Currently it works with multiple values separated by a whitespace, like in this [example from another community](https://github.com/inab/TCGA_benchmarking_workflow/blob/master/nextflow.config#L54):
```
  challenges_ids  = "ACC BRCA"
```

summary_workflows/quantification/main.nf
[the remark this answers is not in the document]
dominikburri: Yes, I think the changes you mention make sense and are sufficient.

Thanks for the explanation. I think this answers my question and I think it is a good approach to overwrite defaults. We only need to make sure we are aware of them.

dominikburri:
Thank you @AsierGonzalez for the explanations. I think the subdirectories and how you explain it makes sense. 

### Update

I could now test on CentOS 7.  Without any modifications `nextflow run main.nf -profile docker` runs without problems. But it is using the Docker images on apaeval.
So I built the images with `quantification_dockers/build.sh` and used the local images by specifying them in `nextflow.config`. This also worked. 

I could not go beyond the test data yet, e.g. using multiple participants, but I intend to do so.

summary_workflows/quantification/main.nf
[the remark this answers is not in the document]
AsierGonzalez: I updated the [README](https://github.com/iRNA-COSI/APAeval/blob/add_working_Q2_summaryworkflow/summary_workflows/quantification/README.md) and [main.nf](https://github.com/iRNA-COSI/APAeval/blob/add_working_Q2_summaryworkflow/summary_workflows/quantification/main.nf#L15) as we agreed.

summary_workflows/quantification/quantification_dockers/q2_metrics/Dockerfile
```diff
@@ -16,5 +16,3 @@ RUN apt-get update && apt-get -y install procps && rm -rf /var/lib/apt/lists/* &
```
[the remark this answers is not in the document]
yuukiiwa: got that removed

summary_workflows/quantification/README.md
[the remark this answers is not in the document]
yuukiiwa: I have included the two methods to build docker images (ask users to choose one) in both of the `README.mds`

summary_workflows/quantification/quantification_dockers/README.md
yuukiiwa: I have added a note here: (note: the `tag_id` should match the one in ../`nextflow.config`)

summary_workflows/quantification/quantification_dockers/q2_metrics/compute_metrics.py
yuukiiwa: I have added `window` as an argument (default: 15) to optionally pass in different window size

dominikburri: Thanks @yuukiiwa , I added a commit b5773862f9c2bd8301da76c14af76bb1db89ecbe such that the window parameter is passed through.

dominikburri:
Thank you @AsierGonzalez  and @yuukiiwa for the changes! 
I tested the workflow and had to make some changes such that the window parameter is correctly passed. Now it works like a charm :)
The only thing I'm not sure about is whether to keep having the apaeval containers or to change it to local containers. For testing purposes, I did this manually, so I can easily commit this change as well. 
I guess using local ones is better (eventhough the test will not work when no images are built). Opinions @yuukiiwa or @AsierGonzalez ?

dominikburri:
> Thank you @AsierGonzalez and @yuukiiwa for the changes! I tested the workflow and had to make some changes such that the window parameter is correctly passed. Now it works like a charm :) The only thing I'm not sure about is whether to keep having the apaeval containers or to change it to local containers. For testing purposes, I did this manually, so I can easily commit this change as well. I guess using local ones is better (eventhough the test will not work when no images are built). Opinions @yuukiiwa or @AsierGonzalez ?

Oh I just realized that having the window size does not make much sense to have in `nextflow.config` and `main.nf`, or does it? I mean the window parameter has to be passed into the python script and this should be done within nextflow no? but then it is specific for Q2 and the window parameter does not make sense for other quantification challenges.

AsierGonzalez:
Thank you for the improvements @dominikburri!

 > The only thing I'm not sure about is whether to keep having the apaeval containers or to change it to local containers. For testing purposes, I did this manually, so I can easily commit this change as well.
I guess using local ones is better (eventhough the test will not work when no images are built). Opinions @yuukiiwa or @AsierGonzalez ?

In OpenEBench we prefer to use remote images since it improves reproducibility so let's keep them.

dominikburri:
> I don't understand the first part. If you need to specify the window size it is fine as you have done it. If this only applies to Q2 I don't think it's such a big deal, the real question is whether we want to be able to tune the value of the window. If yes, I think it's good like this.

I think the window parameter only applies to Q2. And I think we want to be able to tune it (at least have the option to).
Now, if we use this as template, then the window parameter should ideally not be propagated as it might be confusing. But then again it is not very complicated so I think it is okay like that. 

I guess we still need two approving reviews.

AsierGonzalez:
I agree.

Regarding the approving reviews, I still need to change the plot. Sorry I've been busy debugging something else. I'll let you know when  it's ready

yuukiiwa:
Thanks, @AsierGonzalez!

summary_workflows/quantification/quantification_dockers/q2_metrics/compute_metrics.py
```diff
@@ -54,33 +53,30 @@ def compute_metrics(input_participant, gold_standards_dir, challenge_types, part
...
 #                                       comment="#", header=0, sep='\t',index_col=0)
-        metrics_data=os.path.join(gold_standards_dir, "siControl_R2.MACEseq.mm10.bed") ##PLEASE CHANGE THIS LATER
+        metrics_data=os.path.join(gold_standards_dir, challenge + ".bed") ##NO LONGER HARD-CODED
 
         # metric on the number of matched sites
-        window = 15
+        #window = 15
         match_with_gt_run = Leo_MatchPAsites_GT.match_wtih_gt(input_participant,metrics_data,window)
         merged_bed_df, n_matched_sites, n_unmatched_sites = match_with_gt_run[0], match_with_gt_run[1], match_with_gt_run[2]
+        percent_matched=n_matched_sites/(n_matched_sites+n_unmatched_sites)
```
ninsch3000, line 62: This metric is never mentioned in the Q2 specification (there, only correlation is mentioned, also in the specification that has been updated in the meanwhile by @daneckaw (https://github.com/iRNA-COSI/APAeval/pull/210/files#diff-752c23db2dc0d0c543326d5e5ca8bb7d288d17f8b2c70a7388b8332c585d6d7f)).    
Either this has to be added to the specification, or removed here. I know it appears a couple of times downstream, as the OEB .jsons for the plots also specify this "percent_matched" as a metric.

summary_workflows/quantification/Q2_benchmark/Q2_benchmark_specification.md
```diff
@@ -39,59 +39,43 @@ To compute the metrics, the poly(A) sites identified from the RNAseq data using
...
  
  The following table lists the attribute names, value types and units, and a
 description of each attribute-value pair:
  
- | Attribute | Type | Unit | Description |
- | --- | --- | --- | --- |
- | `correlation_coefficient` | `float` | N/A | Correlation between RNAseq-based quantification and 3'end-seq quantification |
+| Attribute | Type | Unit | Description |
+| --- | --- | --- | --- |
+| `correlation_coefficient` | `float` | N/A | Correlation between RNAseq-based quantification and 3'end-seq quantification |
```
ninsch3000, line 73: This file is also changed in another PR (https://github.com/iRNA-COSI/APAeval/pull/210/files#diff-752c23db2dc0d0c543326d5e5ca8bb7d288d17f8b2c70a7388b8332c585d6d7f) which will most certainly lead to conflicts, as the changes there are different from those here. Which are the "true" specifications?

summary_workflows/quantification/quantification_dockers/q2_validation/validation.py
```diff
@@ -67,7 +76,7 @@ def  validate_input_data(input_participant,  public_ref_dir, community, challeng
                 methods_check = list(public_ref_data.iloc[:, 0].values)
```
ninsch3000, line 76: The file public_ref_data is never used for validation. Instead the validation checks (apparently checking whether the input file format is valid bed) are coded here in this script.    
I find this slightly confusing, can it be cleaned up? E.g. remove all parts about the public_ref, especially the for loop we are in here? The checks performed here have nothing to do with the files present in public_ref_dir. But then again, the OEB schema might be unhappy if we don't use a public_ref file, isn't it? Any thoughts @AsierGonzalez ?

summary_workflows/quantification/quantification_dockers/q2_validation/validation.py
```diff
@@ -2,17 +2,18 @@
...
 
 parser = ArgumentParser()
-parser.add_argument("-i", "--participant_data", help="list of cancer genes prediction", required=True)
+parser.add_argument("-i", "--participant_data", help="execution workflow prediction outputs", required=True)
 parser.add_argument("-com", "--community_name", help="name of benchmarking community", required=True)
-parser.add_argument("-c", "--challenge_types", nargs='+', help="list of types of cancer selected by the user, separated by spaces", required=True)
+parser.add_argument("-c", "--challenge_types", nargs='+', help="list challenges selected by the user, separated by spaces", required=True)
 parser.add_argument("-p", "--participant_name", help="name of the tool used for prediction", required=True)
-parser.add_argument("-r", "--public_ref_dir", help="directory with the list of cancer genes used to validate the predictions", required=True)
+parser.add_argument("-r", "--public_ref_dir", help="directory with the list of PAS ground truth used to validate the predictions", required=True)
```
ninsch3000, line 16: This directory is never used, see comment below. Furthermore, is it possible that the help message is not quite right? If anything, shouldn't it be about valid file formats and valid values within the files?

summary_workflows/quantification/README.md
```diff
@@ -1,41 +1,46 @@
...
-python3 match_with_gt.py adultCortex.PAPERCLIP.mm10.bed siControl_R1.MACEseq.mm10.bed 15
+# OpenEBench assessment workflow in Nextflow
+
+Example pipeline with Nextflow used to assess results, comparing the metrics being computed with this workflow with APAeval pilot benchmark results.
+
+#### DON'T FREAK OUT IF YOU'RE UNFAMILIAR WITH `NEXTFLOW`! MOST CHANGES YOU'LL MAKE ARE IN `PYTHON`! 😉
+There are three steps in the summary workflow:
+ - Validation
+    - `input_file`: tab-separated output file from execution workflow
+    - Change the `benchmarking_dockers/apaeval_validation/validation.py` for the specific input_file
```
ninsch3000, line 9: Maybe "for proper validation of the files belonging to different benchmarking events"

summary_workflows/quantification/README.md
```diff
@@ -1,41 +1,46 @@
...
+    - Each input_file may have different fields from different execution workflows
```
ninsch3000, line 10: Don't understand

summary_workflows/quantification/README.md
```diff
@@ -1,41 +1,46 @@
...
+    - `public_ref/[validation_ref].txt` stores the values required to be in the input_files for validating the input_file
```
ninsch3000, line 11: I think it has to be [validadion_ref].[extension], as it doesn't have to be text. And again the description doesn't seem to be quite precise (see other comment within validation.py)

summary_workflows/quantification/README.md
```diff
@@ -1,41 +1,46 @@
...
+There are three steps in the summary workflow:
+ - Validation
+    - `input_file`: tab-separated output file from execution workflow
+    - Change the `benchmarking_dockers/apaeval_validation/validation.py` for the specific input_file
+    - Each input_file may have different fields from different execution workflows
+    - `public_ref/[validation_ref].txt` stores the values required to be in the input_files for validating the input_file 
+    - The `[output].json` file is not used in the subsequent steps
+ - Metrics Computation
+    - `input_file`: tab-separated output file from execution workflow
+    - Change the `benchmarking_dockers/apaeval_metrics/compute_metrics.py` for the specific input_file and the specific metric calculation
```
ninsch3000, line 15: " for the specific input file *type*" maybe?

summary_workflows/quantification/main.nf
```diff
@@ -3,23 +3,26 @@
...
- 	    nextflow run main.nf -profile docker --input {driver.genes.file} --public_ref_dir {validation.reference.file} --participant_id {tool.name} --goldstandard_dir {gold.standards.dir} --cancer_types {analyzed.cancer.types} --assess_dir {benchmark.data.dir} --results_dir {output.dir}
+ 	    nextflow run main.nf -profile docker --input {execution.wf.APA.prediction.file} --public_ref_dir {validation.reference.file} --participant_id {tool.name} --goldstandard_dir {gold.standards.dir} --challenges_ids {analyzed.challenges} --assess_dir {benchmark.data.dir} --results_dir {output.dir}
 	    Mandatory arguments:
-                --input		List of cancer genes prediction
-				--community_id			Name or OEB permanent ID for the benchmarking community
-                --public_ref_dir 		Directory with list of cancer genes used to validate the predictions
+                --input		BED/TXT file with APA site information
+		--community_id			Name or OEB permanent ID for the benchmarking community
+                --public_ref_dir 		Directory with external reference file(s) for validating the structure of the output file(s) from the execution workflows (not utilized currently)
```
ninsch3000, line 22: This description sounds better than the ones in README and validation.py

summary_workflows/quantification/nextflow.config
```diff
@@ -68,7 +68,10 @@ params  {
   outdir = "sample_out/results"
```
ninsch3000, line 68: outdir is defined twice (see two line above)

ninsch3000 (commented):
Thanks a lot @yuukiiwa @AsierGonzalez @dominikburri @daneckaw for all the work on this PR. I have to admit that I still don't understand the whole summary wf, especially the nested .json outputs, and the fact whether we compute one or several metrics per summary workflow.   
However, I managed to run the workflow locally on my laptop by following the README, yay!! And I wrote some rather random and unconnected comments to different files, anything but a complete review, sorry.

ninsch3000:
Actually, according to the schema here: https://github.com/iRNA-COSI/APAeval/tree/main/summary_workflows we would indeed compute several metrics within one summary workflow run. The swfs would of course differ between benchmarking events (Q, I, D), but between challenges within the events only *if needed* (e.g. if there's a challenge with a dataset on which particular metrics don't or do make sense). I'm hoping @AsierGonzalez is saying the same thing [here](https://github.com/iRNA-COSI/APAeval/pull/182#issuecomment-1030209659) and @dominikburri  was mixing up challenges and metrics at some point ;)

daneckaw:
I updated the specifications here #210 
This hopefully clarifies some things about how our benchmarks map to OEB challenges and how I think we might want to use the `window` parameter.

> This file is also changed in another PR (https://github.com/iRNA-COSI/APAeval/pull/210/files#diff-752c23db2dc0d0c543326d5e5ca8bb7d288d17f8b2c70a7388b8332c585d6d7f) which will most certainly lead to conflicts, as the changes there are different from those here. Which are the "true" specifications?

@ninsch3000 I think we should stick to editing Q2 specifications in #210 and I think matched/unmatched sites belong more to identification metrics than quantification.

summary_workflows/quantification/quantification_dockers/q2_metrics/compute_metrics.py
AsierGonzalez: You are right, we said we would only output the correlation, we need to fix this.

summary_workflows/quantification/Q2_benchmark/Q2_benchmark_specification.md
AsierGonzalez: The file in this repo was last updated in November 2021 and, as you say, @daneckaw has been working on the specifications since then so we should give priority to the content in her PR #210.

summary_workflows/quantification/quantification_dockers/q2_validation/validation.py
AsierGonzalez: As we discussed on Slack, we can safely remove all references to `public_ref_data`, related variables and the the file [public_ref.txt](https://github.com/iRNA-COSI/APAeval/blob/add_working_Q2_summaryworkflow/summary_workflows/quantification/APAeval_Q2_test_data/public_ref/public_ref.txt) itself.

summary_workflows/quantification/README.md
AsierGonzalez: Given that this README belongs to a specific benchmarking event I think we should update this to explain what we validation actually does, i.e. checks that the BED file has the correct format.
We can have one README one level up that explains this conceptually if you want it.

summary_workflows/quantification/README.md
AsierGonzalez: Uuuum, it seems to suggest that different tools could produce different BED formats but I don't think this is the case since you made sure that they would be compatible, right? If so, this can be removed

summary_workflows/quantification/README.md
AsierGonzalez: You are right.
Anyway, as we said, all reference to the reference file can be removed.

summary_workflows/quantification/README.md
AsierGonzalez: To be honest, I don't see anything wrong with the use of "input_file" here given that it's defined in the previous row. If anything, I would format it somehow to emphasise it, for example using `code` or _italic_ styles. However, I would do this every time "input_file" appears in the README.

Other than that, I would prefer if we rewrote the README to explain what each step does specifically and here it said that we are measuring the correlation. We can do that when we finish implementing it.

summary_workflows/quantification/main.nf
AsierGonzalez: Yep, but this also can go now

summary_workflows/quantification/quantification_dockers/q_consolidation/manage_assessment_data.py
```diff
@@ -1,30 +1,114 @@
...
+import sys
+import datetime
 from argparse import ArgumentParser
-import numpy as np
 from assessment_chart import assessment_chart
 
-DEFAULT_eventMark = '2018-04-05'
+DEFAULT_OEB_API = "https://dev-openebench.bsc.es/api/scientific/graphql"
+DEFAULT_eventMark_id = "OEBE0070000000"
+METRICS =  {"correlation":"OEBM0070000000"} ## is this needed?
```
ninsch3000, line 15: Not sure whether METRICS is needed, as the parts of the code using it have been removed. I'll comment on the positions where METRICS is used in the TCGA workflow

summary_workflows/quantification/quantification_dockers/q_consolidation/manage_assessment_data.py
```diff
@@ -80,7 +166,7 @@ def generate_manifest(data_dir,output_dir, participant_data):
                 metric_Y = "precision"
```
ninsch3000, line 166: This part (line 163-166) has to be updated, but I'm not sure how. We have only one metric "correlation". What will be the other axis (if this means we need a different type of plot, we'd have to change that too. But I don't know where)

summary_workflows/quantification/quantification_dockers/q_consolidation/manage_assessment_data.py
```diff
@@ -80,7 +166,7 @@ def generate_manifest(data_dir,output_dir, participant_data):
...
             
             aggregation_file = {
-                "_id": "TCGA:{}_{}_Aggregation".format(DEFAULT_eventMark, challenge),
+                "_id": "{}:{}_{}_Aggregation".format(metrics_file[0]["community_id"],event_date, challenge),
                 "challenge_ids": [
                     challenge
                 ],
```
ninsch3000, line 172: after line 175 (after challenge participants), the TCGA script contains the following:
```python
"metrics":{
    "metric_x_id": METRICS['TPR'],
    "metric_y_id": METRICS['precision']
},
```
That's one place where METRICS is used. Is it needed? Can't find key `metric_x_id` anywhere

summary_workflows/quantification/quantification_dockers/q_consolidation/manage_assessment_data.py
```diff
@@ -121,7 +207,10 @@ def generate_manifest(data_dir,output_dir, participant_data):
             elif metrics_data["metrics"]["metric_id"] == metric_Y:
                 new_participant_data["metric_y"] = metrics_data["metrics"]["value"]
 
-        # copy the assessment files to output directory
+        new_participant_data["participant_id"] = participant_id
+        aggregation_file["datalink"]["inline_data"]["challenge_participants"].append(new_participant_data)
+
```
ninsch3000, line 212: Here again in TCGA METRICS is used:
```python
metrics = {
    "metric_x_id": METRICS['TPR'],
    "metric_y_id": METRICS['precision']
}
aggregation_file["datalink"]["inline_data"]["metrics"] = metrics
```
Needed or not?

summary_workflows/quantification/quantification_dockers/q_consolidation/manage_assessment_data.py
```diff
@@ -121,7 +207,10 @@ def generate_manifest(data_dir,output_dir, participant_data):
             elif metrics_data["metrics"]["metric_id"] == metric_Y:
                 new_participant_data["metric_y"] = metrics_data["metrics"]["value"]
 
-        # copy the assessment files to output directory
+        new_participant_data["participant_id"] = participant_id
+        aggregation_file["datalink"]["inline_data"]["challenge_participants"].append(new_participant_data)
+        
+        # copy the assessment file of the current participant to output directory
```
ninsch3000, line 213: Why are we doing this?

summary_workflows/quantification/quantification_dockers/q_consolidation/manage_assessment_data.py
```diff
@@ -121,7 +207,10 @@ def generate_manifest(data_dir,output_dir, participant_data):
             elif metrics_data["metrics"]["metric_id"] == metric_Y:
```
ninsch3000, line 207: lines 186-198 I couldn't figure out. Don't know which file is read here and whether it will ever exist, I couldn't find the key `challenge_type` anywhere.

ninsch3000 (commented):
Hi @AsierGonzalez 
I included the two DB query functions that are present in the TCGA workflow in `manage_assessment_data.py`. Had to make a small change to handle the parameter `offline`.  These OEB DB calls do work on my computer.   
I also made some comments on that script where I saw differences to the TCGA version, and where I don't understand the code. It's mainly about whether we need to handle the OEB metrics identifier, or whether we can completely omit that (or maybe it is handled somewhere else which I didn't see).    
I still don't understand how all the `.json` files are connected and where which is exactly written. Another thing I noticed is that the nextflow logs for the individual steps do show some error messages (Like
```
# validation:
OS error: [Errno 2] No such file or directory: ''
Could not create output path: validation.json

# metrics:
OS error: [Errno 2] No such file or directory: ''
Could not create output path: assessment.json

# consolidation:
OS error: [Errno 2] No such file or directory: ''
Could not create output path: data_model_export.json
```
The workflow runs through though and nextflow doesn't complain at all, but it makes me wonder whether this is intended behaviour. Those filenames are the ones defined in `main.nf`, while we end up with files named according to the `nextflow.config`. Well, can't explain it better...

Would be happy about your feedback and if you could let me know where else I can do here.

summary_workflows/quantification/main.nf
```diff
@@ -0,0 +1,163 @@
...
+	    *1 iRNA COSI
+	    *2 Genomic Institute of Singapore, A*STAR, Singapore
+	    *3 Barcelona Supercomputing Center, Barcelone, Spain
+	    *4 Biozentrum, University of Basel, Switzerland
+	    =====================================================
+	    Usage:
+	    Run the pipeline with default parameters read from nextflow.config:
+	    nextflow run main.nf -profile docker
+	    Run with user parameters:
+	   nextflow run main.nf -profile docker --input {execution.wf.APA.prediction.file} --participant_id {tool.name} --goldstandard_dir {gold.standards.dir} --challenge_ids {analyzed.challenges} --assess_dir {benchmark.data.dir} --results_dir {output.dir}
```
dominikburri, line 18: `results_dir` is not described in the mandatoryy arguments sections. 
What is the difference of `assess_dir`and this results directory? Seems to save similar output.

dominikburri (commented):
Hi all, thanks for the latest pushes!
After some discussion with @ninsch3000 I ran some samples from the google drive to check whether is works as intended. The provided test data runs without problems. I hope I use the nomenclature correctly, otherwise please report!
In general it seems to work and from my side it appears ready, however I have some comments and questions:

- it seems I need to manually copy assessment data into the specified directory, such that the aggregation works correctly. Is this intended?
- Running another sample duplicates the participant, is this intended? See example below. 
- When running new challenges, the results seem to be completely removed. 
- Is supplying a list of challenges useful for us? 
- Is this file needed? `summary_workflows/quantification/APAeval_Q_test_data/data/compute_resources.json`
- Is the integration with OEB successful? 

## Example

Note that I changed `nextflow.config` a bit, such that I have a new variable `results_dir` that is the top directory of the other ones:
```
  results_dir = "output"
  validation_result = "${params.results_dir}/validated_participant_data.json"
  assessment_results = "${params.results_dir}/assessment_datasets.json"
  outdir = "${params.results_dir}/results"
  statsdir = "${params.results_dir}/stats"
  consolidation_result = "${params.results_dir}/consolidated_result.json"
  otherdir = "${params.results_dir}/other"
```
As @ninsch3000 is working on this I will not commit this as changes.

### samples
The samples I obtained with @ninsch3000 from google drive and they are very likely not right, as they seem to be outdated. Anyway, for the purpose of testing this workflow, the samples just need to have some values and be able to report some results. So don't take the metrics (percent matched and correlation) as the final results.
In the example below, only `Mayr_M_R2.SRR6795691.3seq.hg38.bed` is used as challenge.
Also, needed to switch columns 4 and 5 in labrat samples (done manually).
```
Mayr/
├── labrat
│   ├── Mayr_M_R2_1_labrat_quant._mapped.bed
│   └── Mayr_M_R6_1_labrat_quant._mapped.bed
├── Mayr_ground_truth
│   ├── Mayr_M_R1.SRR6795690.3seq.hg38.bed
│   └── Mayr_M_R2.SRR6795691.3seq.hg38.bed
└── qapa
    ├── Mayr_M_R2.qapaFixed.bed
    └── Mayr_M_R6.qapaFixed.bed
```
### run commands
Correctly running the LABRAT and QAPA output. That is, R2 output is matched to R2 challenge.
```bash
nextflow run main.nf -profile docker \
    --input Mayr/labrat/Mayr_M_R2_1_labrat_quant._mapped.bed \
    --participant_id "LABRAT" \
    --goldstandard_dir Mayr/Mayr_ground_truth \
    --challenge_ids "Mayr_M_R2.SRR6795691.3seq.hg38" \
    --assess_dir Mayr/assess \
    --results_dir Mayr/output;
mv Mayr/output/results/Mayr_M_R2.SRR6795691.3seq.hg38/Mayr_M_R2.SRR6795691.3seq.hg38.json Mayr/assess;
nextflow run main.nf -profile docker \
    --input Mayr/qapa/Mayr_M_R2.qapaFixed.bed \
    --participant_id "QAPA" \
    --goldstandard_dir Mayr/Mayr_ground_truth \
    --challenge_ids "Mayr_M_R2.SRR6795691.3seq.hg38" \
    --assess_dir Mayr/assess \
    --results_dir Mayr/output;
mv Mayr/output/results/Mayr_M_R2.SRR6795691.3seq.hg38/Mayr_M_R2.SRR6795691.3seq.hg38.json Mayr/assess;
```
then, incorrectly running R6 samples on the same R2 challenge. I mean it does not make sense to use different samples on the same challenge, but I was curious what would happen.
```bash
nextflow run main.nf -profile docker \
    --input Mayr/labrat/Mayr_M_R6_1_labrat_quant._mapped.bed \
    --participant_id "LABRAT" \
    --goldstandard_dir Mayr/Mayr_ground_truth \
    --challenge_ids "Mayr_M_R2.SRR6795691.3seq.hg38" \
    --assess_dir Mayr/assess \
    --results_dir Mayr/output;
mv Mayr/output/results/Mayr_M_R2.SRR6795691.3seq.hg38/Mayr_M_R2.SRR6795691.3seq.hg38.json Mayr/assess;
nextflow run main.nf -profile docker \
    --input Mayr/qapa/Mayr_M_R6.qapaFixed.bed \
    --participant_id "QAPA" \
    --goldstandard_dir Mayr/Mayr_ground_truth \
    --challenge_ids "Mayr_M_R2.SRR6795691.3seq.hg38" \
    --assess_dir Mayr/assess \
    --results_dir Mayr/output;
mv Mayr/output/results/Mayr_M_R2.SRR6795691.3seq.hg38/Mayr_M_R2.SRR6795691.3seq.hg38.json Mayr/assess;
```

### results
running above gives this `Mayr/assess/Mayr_M_R2.SRR6795691.3seq.hg38/Mayr_M_R2.SRR6795691.3seq.hg38.json`
```json
{
    "_id": "APAeval:2022-03-25_Mayr_M_R2.SRR6795691.3seq.hg38_Aggregation",
    "challenge_ids": [
        "Mayr_M_R2.SRR6795691.3seq.hg38"
    ],
    "datalink": {
        "inline_data": {
            "challenge_participants": [
                {
                    "metric_x": 0.2218817039534171,
                    "metric_y": 0.45867288632847986,
                    "participant_id": "LABRAT"
                },
                {
                    "metric_x": 0.2742682394058541,
                    "metric_y": 0.32025646910248257,
                    "participant_id": "QAPA"
                },
                {
                    "metric_x": 0.22108208955223882,
                    "metric_y": 0.5137723632936275,
                    "participant_id": "LABRAT"
                },
                {
                    "metric_x": 0.2759898227915904,
                    "metric_y": 0.3866288173952344,
                    "participant_id": "QAPA"
                }
            ],
            "visualization": {
                "type": "2D-plot",
                "x_axis": "percent_matched",
                "y_axis": "correlation"
            }
        }
    },
    "type": "aggregation"
}
```
### notes
- Does the duplication of `participant_id` cause problems? I know it is unlikely to happen, so @AsierGonzalez can probably tell me that it is unnecessary to consider this case.
- The generated svg's contain all 4 participants.
- The svg seems to differentiate based on color and shape. What is the maximum number of participants we can plot? I guess we won't run into problems as we likely have not more than 10 participants per challenge.

ninsch3000:
> * It seems I need to manually copy assessment data into the specified directory, such that the aggregation works correctly. Is this intended?

This is taken care of in the latest commit, the aggregation file is written to the data directory in case the swf is run offline

ninsch3000:
> * Is this file needed? `summary_workflows/quantification/APAeval_Q_test_data/data/compute_resources.json`

No

ninsch3000:
> * The svg seems to differentiate based on color and shape. What is the maximum number of participants we can plot? I guess we won't run into problems as we likely have not more than 10 participants per challenge.

This is implemented in `assessment_chart.py`. There are 25 different shapes

ninsch3000:
> incorrectly running R6 samples on the same R2 challenge.

Good point. We have to be aware that it's up to the user to provide the correct samples for a challenge, and that the filenames will not be checked. If we are consistent in naming our execution workflow outputs such a check might be worth implementing though?

ninsch3000 (approved):
merge now, create new issues and new PRs for specific tasks

dominikburri (approved):
Approving for now with the possibility of amending and adding changes on top of this PR.

AsierGonzalez (approved):

AsierGonzalez:
Ok, let's go for it. I run it locally with the default config and it worked fine.

I'll still add answers to your comments since you made some very good points that are worth answering and, probably, solving in another branch+PR.

Great job @ninsch3000, @dominikburri ,  @yuukiiwa! :raised_hands: :raised_hands: :raised_hands:

summary_workflows/quantification/quantification_dockers/q_consolidation/manage_assessment_data.py
AsierGonzalez: I'm checking this with a colleague but it seems that the `METRICS` dictionary and the `.datalink.inline_data.metrics` object in the aggregation file were added so that the metric id could be shown on the plots. I'll confirm on Tuesday but I'm rather sure that all the references to `METRICS` can be removed. Even more so considering that you commented those parts out and it kept working.

summary_workflows/quantification/quantification_dockers/q_consolidation/manage_assessment_data.py
AsierGonzalez: I suggest we remove this code as I don't think it ever runs with the current directory structure. And even if it ran, it seems that the content added to the `challenge_participants` list in [rows 194-198](https://github.com/iRNA-COSI/APAeval/blob/07b96d7f013a4aa0241bc4ad35772329b02cb92e/summary_workflows/quantification/quantification_dockers/q_consolidation/manage_assessment_data.py#L194-LL198) would never make it into the aggregation file since I don't see it being referenced again.

AsierGonzalez: Correction to myself: the `challenge_participants.append()` in row 194 would actually change the content of `challenge_participants` in `aggregation_file` since arrays are passed as reference.

AsierGonzalez: Anyway, I still think we can ignore it since that doesn't change what I said about this snippet not  running due to the current directory structure.

summary_workflows/quantification/quantification_dockers/q_consolidation/manage_assessment_data.py
AsierGonzalez: I'm afraid I don't have an answer to this question. This snippet copies the assessment objects into a file in the `results` directory, which is used by the VRE. Unfortunately, I cannot tell whether or not this file is needed by the VRE, so I think it is best to just keep it.

summary_workflows/quantification/quantification_dockers/q_consolidation/manage_assessment_data.py
AsierGonzalez: I can confirm that all references to `METRICS` can be removed, there's no need for the mapping for now. It was just some new feature that a colleague was testing.

summary_workflows/quantification/quantification_dockers/q_consolidation/manage_assessment_data.py
AsierGonzalez: We don't need `metrics` or any of its children (`metric_x_id`, `metric_y_id`), this was related to the test that a colleague did that I mentioned [above](https://github.com/iRNA-COSI/APAeval/pull/182#discussion_r852898176).

PR #198, opened 2021-11-24
feat: PAQR execution workflow

### LINKED ISSUE #41: Execution Workflow: PAQR
## WHAT    
**Write** execution workflow for [PAQR](https://github.com/zavolanlab/PAQR_KAPAC). Use the provided [small files for testing](https://github.com/iRNA-COSI/APAeval/tree/main/tests/test_data) (running the workflow on real data is a different issue).

## CHECKLIST   
* [ ] Comment your code

* [ ] ~Output: Adhere to output specification for Identification challenge~ No identification

    > - gene ID

ninsch3000:
Fixes #41 #223

## TODO
- preprocessing
- postprocessing
- creation or download of poly(A) sites reference file?
- cluster configs
- paqr module needs samples.tsv (instead of samples.csv). Can this be overridden somehow?

## Checklist

- [x] I have tested my code


dominikburri (commented):
Hi @ninsch3000, thanks for the great work!
I recognize that this is still a draft but I've got some comments.

* I could not go through the test, as I got an "name 'github' is not defined" error. Not sure what the problem is, but seems like a missing import or so.
* I think we decided not to have the subdirectories such as `PAQR_snakemake`, thus I think it can be deleted. 
* Include in the readme how to run the workflow, and that the current configuration is set up for the test data.
* the config file seems to have parameters with similar meaning, e.g. samples table and out directory specification. I marked those I found.
* some comments regarding left-overs from the template

execution_workflows/PAQR/PAQR_snakemake/workflow/rules/paqr.smk
```diff
@@ -0,0 +1,25 @@
...
+        HTTP.remote("raw.githubusercontent.com/zavolanlab/PAQR2/bd9781f1420ba39648120a6898038f49728541f3/scripts/{script}", keep_local=True)
+    output:
+        os.path.join(config["scriptsdir"], "{script}")
+    shell:
+        "mv {input} {output}"
+
+
+# Import the PAQR module
+module paqr:
+    snakefile: github("zavolanlab/PAQR2", path="Snakefile", commit="fdf56bc")
```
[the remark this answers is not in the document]
ninsch3000: I updated the conda env file in this PR, did you see that [here](https://github.com/iRNA-COSI/APAeval/blob/4d15f17b85ee18af4ffb0cfed5b42a5e3537a92e/apaeval_env.yaml). Could you double check whether you have snakemake 10? Maybe this will already resolve the issue.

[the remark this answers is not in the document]
dominikburri: You're right, I missed that. I updated the conda environment by removing and reinstalling the environment and now have snakemake version 6.10.0.
Now I don't get this message anymore.

dominikburri:
> Hi @ninsch3000, thanks for the great work! I recognize that this is still a draft but I've got some comments.
> 
> * I could not go through the test, as I got an "name 'github' is not defined" error. Not sure what the problem is, but seems like a missing import or so.
> * I think we decided not to have the subdirectories such as `PAQR_snakemake`, thus I think it can be deleted.
> * Include in the readme how to run the workflow, and that the current configuration is set up for the test data.
> * the config file seems to have parameters with similar meaning, e.g. samples table and out directory specification. I marked those I found.
> * some comments regarding left-overs from the template

**UPDATE**: actually using the bumped snakemake version helped and I don't get the "name 'github' is not defined" error anymore. 

Running the test`dryrun.sh` works, but I get an error with `run_local.sh`:
`ERROR  : Failed to create user namespace: user namespace disabled`
which could be due to singularity.
Manually removing this from the `apaeval_execution_workflows` environment with `conda remove singularity` and instead using my locally installed Singularity (v3.5.2), the test successfully finishes. 

So, maybe we need to delete the conda-based singularity (`conda-forge::singularity=3.7.1`) in `apaeval_env.yaml`? Maybe it's just not working for me because I don't have root access on my machine.

apaeval_env.yaml
```diff
@@ -5,7 +5,7 @@ channels:
   - defaults
 dependencies:
   - nextflow=21.04.0
-  - snakemake=6.4.0
+  - snakemake=6.10.0
```
faricazjj, line 8: I think this change should be a separate issue of updating apaeval env snakemake version and separate PR from PAQR execution workflow so that it's easier to find and revert, plus it's not PAQR specific :D

execution_workflows/PAQR/PAQR_snakemake/README.md
```diff
@@ -0,0 +1,81 @@
...
+>     * Test your code with [snakemake --lint](https://snakemake.readthedocs.io/en/stable/snakefiles/writing_snakefiles.html#best-practices).
+>     * One (shell) command per rule. 
+>     * If samples differ in a meaningful way (e.g. single end and paired end samples), it might be better to write subworkflows within `workflow/rules`. 
+>     * If applicable, each rule has it's own [Docker container](https://www.docker.com/resources/what-container) or conda environment.
+> * There are some shell scripts which can be used to start a snakemake run. Adjust the name of the `config file` in these scripts:
+>     * `dryrun.sh`
+>     * `rulegraph.sh` (Also adjust name of output `.png`)
+>     * `run_local.sh`
+> * Check out the pilot benchmark at `tests/pilot_benchmark/snakemake` for a running example. It illustrates how the execution workflow can look in practice. 
+> * Adjust this `README.md`: Delete this description section and populate the sections below with your awesome experience ;)
```
faricazjj, line 22: can be deleted

execution_workflows/PAQR/PAQR_snakemake/README.md
```diff
@@ -0,0 +1,81 @@
...
+- `PAQ_coverage_unstranded`: "no" for stranded data
+- `PAQ_read_length`: avg read length of the samples
+
+### Annotation dependent
+- `biotype_key`: "transcript_biotype" if using ensembl annotations, "transcript_type" if using gencode
+
+
+## Output & post-processing
+
+{Describe output files and postprocessing steps if necessary.}
```
faricazjj, line 73: this section is not updated yet

execution_workflows/PAQR/PAQR_snakemake/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Notes
+
+{Notes about the METHOD.
```
faricazjj, line 77: this section is not updated yet, could also add author contacts

execution_workflows/PAQR/PAQR_snakemake/config/config.PAQR.yaml
```diff
@@ -0,0 +1,151 @@
...
+
+
+
+
+
+
+
+
+
+
```
faricazjj, line 145: formatting: remove empty lines

execution_workflows/PAQR/PAQR_snakemake/README.md
```diff
@@ -0,0 +1,81 @@
...
+## Input & pre-processing
+> IMPORTANT:   
+PAQR requires samples, annotations and reference poly(A) sites files to all use the same chromosome naming scheme ("chr1" in gencode, "1" in ensembl) or else will fail.
+
+{Describe input files and how they will be processed in order for the method to work. Describe how sample tables have to look like, and any other input that is needed (e.g. genome).}
+
+1. Sample files  
+`.bam` files of all samples have to be provided alongside their respective `.bai` index files (**in the same directory, with the same basename**)
+
+2. Sample table
```
faricazjj, line 40: maybe mention here an example of input sample table is provided in config/samples.tsv

execution_workflows/PAQR/PAQR_snakemake/README.md
```diff
@@ -0,0 +1,81 @@
...
+> IMPORTANT:   
+PAQR requires samples, annotations and reference poly(A) sites files to all use the same chromosome naming scheme ("chr1" in gencode, "1" in ensembl) or else will fail.
+
+{Describe input files and how they will be processed in order for the method to work. Describe how sample tables have to look like, and any other input that is needed (e.g. genome).}
+
+1. Sample files  
+`.bam` files of all samples have to be provided alongside their respective `.bai` index files (**in the same directory, with the same basename**)
+
+2. Sample table  
+A table specifying all samples to be analyzed has to be provided in `.tsv` format (not `.csv`) and contain the following columns:
```
faricazjj, line 41: maybe specify how many rows are required (e.g. at least one row in the table), any constraints as to number of unique conditions (e.g exactly two unique conditions? does it work with one conditions?)

faricazjj (commented):
Thanks for this PR @ninsch3000 ! :D I have some initial formatting comments, will give this another review in a bit!

apaeval_env.yaml
ninsch3000: Agree, I'll create a new issue. Nevertheless, I need to bump snakemake version in this PR, because otherwise the workflow will not run (snakemake modules aren't implemented in 6.4.0)

execution_workflows/PAQR/.gitignore
```diff
@@ -0,0 +1,3 @@
+results*/
+*log*
+.snakemake/
```
dominikburri, line 3: Maybe add the downloaded scripts to gitignore? That is, `workflow/scripts_paqr` and `workflow/scripts_tpas`. 
Though I'm not sure what the best approach is.

dominikburri (dismissed):
Thank you @ninsch3000 for all the work!

The example runs successfully (`bash run_local.sh`).
Can it be that the resulting bed files are empty? I just want to make sure that the test ran completely. 
Otherwise I just have very minor comments. So good to be merged from my side.

ninsch3000:
> Can it be that the resulting bed files are empty? 

Yes, good point! I always assumed that this would be different on real data, but always forgot to test. Just now I ran the workflow on a small C. elegans sample, and there I get some output. However, I only get 8 poly(A) sites, so that's a good reminder that the settings in config are probably not ideal yet. Have to check the publication for their defaults!

ninsch3000:
TO DO:   
- [ ] Check for correct chromosome naming of reference PAS (with or without "chr") and add "chr" if necessary
OR:
- [ ] Remove reference PAS download from workflow and describe it and the correct chr naming necessity in README

SamBryce-Smith (commented):
Great stuff @ninsch3000! My test run completed without errors, although I also got empty output files. Would like to try with a bigger dataset before approving to make sure this is just a consequence of the test data. I don't really have any other changes to suggest.

My thoughts RE handling how to provide reference PolyASite file:

I'd prefer for the file to be provided to the workflow rather than downloaded, mainly for consistency with other workflows (none of which download any reference files). I realise it would be quite easy to run into problems with matching chromosome names as you suggest (PolyASite was processed using Ensembl annotations, we want to use Gencode where possible). 

From memory a while back I thought we decided that we'd just focus on analysing standard chromosomes only (correct me if I'm wrong). If so *adding the 'chr' prefix (outside of the workflow) should be fine*. I don't know whether the sub-setting to standard chroms has been done in practice though.

If we want to convert the chromosome names more robustly, we could take the following approach (and make a utils for it). [This repo](https://github.com/dpryan79/ChromosomeMappings) provides tables mapping chromosome names between Gencode,Ensembl etc. for mouse,human genomes and others. There's also the tool `UpdateContigNames` from [cvbio](https://github.com/clintval/cvbio#updatecontignames) that can take a BED/GFF as input and remap the names given one of these tables. There may not always be direct conversions, but this would at least be more robust than adding 'chr' to everything. Happy to put this together if others agree.

Also, in other workflows we opted against adding the leading 'chr' if absent, although we do check that they are present for tools that hardcode for them (e.g. DaPars). 

Checking the chromosome names match between inputs would be entirely sensible, although I think generally for other workflows we don't do this unless the tools requires a specific naming convention. Maybe we should have done this to begin with (e.g. make a utils), but I think we're a little too late in the project to implement this systematically.

The requirement for the same chromosome names is in a general sense right? By that I mean it's not because PAQR requires a specific naming convention (I'm pretty sure that's not the case), it's just that to match different reference files you need the same names.

ninsch3000:
> The requirement for the same chromosome names is in a general sense right? By that I mean it's not because PAQR requires a specific naming convention (I'm pretty sure that's not the case), it's just that to match different reference files you need the same names.

PAQR requires all inputs (= PolyASite bed, gtf, bam files) to have the same chr naming, else it will not find any matches. But both, with or without "chr" are fine. But yes, PolyASite is without "chr", so we have to adjust

ninsch3000:
Thanks a lot for your thoughts @SamBryce-Smith !   
I think due to lack of time we could really just add "chr" to the PolyASite bed. As I understand from the [Gencode FAQs](https://www.gencodegenes.org/pages/faq.html), Gencode and Ensembl are using the same genome and it's not a problem to insert or delete "chr". The conversion *is* needed between Ensembl and UCSC genomes though. I might be wrong however, because this topic always confuses me :stuck_out_tongue_closed_eyes:    

And I agree, I'll delete the downloading of the PolyASite bed, that didn't feel right anyway.

dominikburri (dismissed):
Thank you @ninsch3000 for the changes!

As we discussed, I added commits to deal with singular PAS, which should increase the number of PAS that PAQR quantifies.

I tested with provided data and it worked.
Also used some human samples, it ran through and the results seem reasonable. 

Good to be merged from my side.

execution_workflows/PAQR/config/config.PAQR.yaml
```diff
@@ -132,12 +121,16 @@ paqr:
...
 
   ############################################
   # Mock params needed for DAG, don't change
-  PAQ_tandem_pas: ""
\ No newline at end of file
+  PAQ_tandem_pas: ""
+  # out- and logdir are set in Snakefile to be subdirectories of the result directory specified at param `out_dir`
+  PAQ_outdir: "results4/paqr"
+  # path for log directory
+  PAQ_logdir: "results4/logs"
```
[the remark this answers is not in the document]
ninsch3000: sorry, it's a remnant from testing. Should be results/logs

SamBryce-Smith (commented):
Thanks @ninsch3000 have completed a run with test data locally! Followed instructions in README to the letter and everything worked perfectly :)

image

Just a quick comment about the quantification BED files. For identification BEDs, we specify these must be [single nucleotide intervals](https://github.com/iRNA-COSI/APAeval/blob/main/execution_workflows/execution_output_specification.md#format-01). Though we don't actually say this as explicitly for [quantification BEDs](https://github.com/iRNA-COSI/APAeval/blob/main/execution_workflows/execution_output_specification.md#format-02) (it's mentioned in the chromStart description), I think it should be the case (and in other PRs I have tried to make sure this is the case).

image

I know it sort of goes against the purpose of why PolyASite defines polyA site clusters, but I worry it would give PAQR an unfair advantage for matching sites to ground truth (the polyA site intervals are already larger to begin with before we add the windows). Perhaps we could take the representative position from the cluster ID/Name as the single-nucleotide interval to represent each cluster?

Happy to discuss final point further and will accept it not being implemented to approve, just think the point needs to be raised regardless. Other comments are very minor

execution_workflows/PAQR/config/config.PAQR.yaml
```diff
@@ -0,0 +1,136 @@
+# Configuration file for snakemake pipeline
+
+# Reference PAS file: must have the same format as PolyASite bed
+# see https://polyasite.unibas.ch/atlas
+
+# reference PAS filename
+ref_PAS_file: "../../tests/test_data/atlas.clusters.2.0.GRCm38.96.Chr_prefix.bed.gz"
```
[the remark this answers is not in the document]
ninsch3000: Judging from the perl script that is gonna load this file, both compressed and uncompressed will work. Will adjust the comment here in the config.

ninsch3000:
Thanks a lot for your review @SamBryce-Smith ! Excellent points, as usual. :star_struck:   

@dominikburri  and me did in fact briefly discuss the issue of the poly(A) site clusters vs. single nuc positions yesterday, but discarded it as "well, that's part of the participants algorithm/mode of function". But you are right of course, the reported intervals are already larger to begin with, which is an "advantage" for matching sites (might be a disadvantage if we're judging precision of location, or penalizing multiple overlaps between one pd and several gt sites, but as I understandm we are not with the quantification metrics. Or are we, @daneckaw ?).   

However, we could then argue, that using PolyASite *clusters* in the first place is an advantage, because other tools might use reference PAS with only single nucleotide positions and thus have a more limited genomic range to even look for poly(A) sites. And so we're back at the old discussion about providing the same reference PAS to all participants, or using the one that the authors intended :)   

That said, I'm nevertheless fine with reporting the representative sites instead of the clusters. But curious to hear @dominikburri 's opinion.

SamBryce-Smith (approved):
Looks good to me thanks for responding to comments @ninsch3000 ! Did another test run and everything seems to be addressed. PAS coordinates are now single nucleotide, PAQR logs end up under the main results directory (`results/logs`), quick note on uncompressed atlas BED being ok have all been addressed. Good to go on my end :sunglasses:

dominikburri (approved):
Hi @ninsch3000, thank you for the changes!

Local test runs without problems.
I also re-ran the human samples we were looking at yesterday, and the results look reasonable.

Good to be merged from my side.

PR #200, opened 2021-12-08
Ewf labrat extension

dominikburri:
This branch includes some fixes and enhancements to run LABRAT.
However, the changes are not complete and some more modifications are desired, see checklist below.

## Checklist

- [x] I have tested my code

### DIFF
--- a/execution_workflows/LABRAT/modules/labrat_calculatepsi.nf
+++ b/execution_workflows/LABRAT/modules/labrat_calculatepsi.nf
@@ -8,6 +8,8 @@ process LABRAT_CALCULATEPSI {
         publishDir "${params.outdir}/labrat", mode: params.publish_dir_mode
         container "quay.io/biocontainers/labrat:0.2.2--pyhdfd78af_0"
 
+        label "process_low"
+
         input:
         val salmon_dir
         path sampcond
--- a/execution_workflows/LABRAT/modules/labrat_calculatepsi.nf
+++ b/execution_workflows/LABRAT/modules/labrat_calculatepsi.nf
@@ -8,12 +8,16 @@ process LABRAT_CALCULATEPSI {
         publishDir "${params.outdir}/labrat", mode: params.publish_dir_mode
         container "quay.io/biocontainers/labrat:0.2.2--pyhdfd78af_0"
 
+        cpus 3
+        memory '16 GB'
+
         input:
         val salmon_dir
         path sampcond
         val conditionA
         val conditionB
         path gff
+        path db
 
         output:
         path "*.psis" , emit: ch_psis
--- a/execution_workflows/LABRAT/modules/labrat_calculatepsi.nf
+++ b/execution_workflows/LABRAT/modules/labrat_calculatepsi.nf
@@ -8,9 +8,6 @@ process LABRAT_CALCULATEPSI {
         publishDir "${params.outdir}/labrat", mode: params.publish_dir_mode
         container "quay.io/biocontainers/labrat:0.2.2--pyhdfd78af_0"
 
-        cpus 3
-        memory '16 GB'
-
         input:
         val salmon_dir
         path sampcond
--- a/execution_workflows/LABRAT/modules/labrat_calculatepsi.nf
+++ b/execution_workflows/LABRAT/modules/labrat_calculatepsi.nf
@@ -18,6 +18,7 @@ process LABRAT_CALCULATEPSI {
         val conditionB
         path gff
         path db
+        val bed
 
         output:
         path "*.psis" , emit: ch_psis
@@ -25,7 +26,8 @@ process LABRAT_CALCULATEPSI {
         path "*.pval" , emit: ch_pval_results
 
         script:
+        def add_to_path = (params.run_quantification) ? "$PWD/${params.outdir}" : ""
         """
-        LABRAT.py --mode calculatepsi --librarytype RNAseq --salmondir $PWD/${params.outdir}/$salmon_dir --sampconds $sampcond --conditionA $conditionA --conditionB $conditionB --gff $gff
+        LABRAT.py --mode calculatepsi --librarytype RNAseq --salmondir $add_to_path/$salmon_dir --sampconds $sampcond --conditionA $conditionA --conditionB $conditionB --gff $gff
         """
 }
--- a/execution_workflows/LABRAT/modules/labrat_maketffasta.nf
+++ b/execution_workflows/LABRAT/modules/labrat_maketffasta.nf
@@ -8,11 +8,14 @@ process LABRAT_MAKETFFASTA {
         publishDir "${params.outdir}/labrat", mode: params.publish_dir_mode
         container "quay.io/biocontainers/labrat:0.2.2--pyhdfd78af_0"
 
+        time '8h'
+
         input:
         tuple path(gff), path(fasta)
 
         output:
         path "*.fasta", emit: ch_tffasta
+        path "*.db", emit: ch_gffutilsdb
 
         script:
         """
--- a/execution_workflows/LABRAT/modules/labrat_maketffasta.nf
+++ b/execution_workflows/LABRAT/modules/labrat_maketffasta.nf
@@ -8,7 +8,7 @@ process LABRAT_MAKETFFASTA {
         publishDir "${params.outdir}/labrat", mode: params.publish_dir_mode
         container "quay.io/biocontainers/labrat:0.2.2--pyhdfd78af_0"
 
-        time '8h'
+        //time '8h'
 
         input:
         tuple path(gff), path(fasta)
--- a/execution_workflows/LABRAT/modules/labrat_maketffasta.nf
+++ b/execution_workflows/LABRAT/modules/labrat_maketffasta.nf
@@ -8,7 +8,7 @@ process LABRAT_MAKETFFASTA {
         publishDir "${params.outdir}/labrat", mode: params.publish_dir_mode
         container "quay.io/biocontainers/labrat:0.2.2--pyhdfd78af_0"
 
-        //time '8h'
+        label "process_long"
 
         input:
         tuple path(gff), path(fasta)
--- a/execution_workflows/LABRAT/modules/salmon_index.nf
+++ b/execution_workflows/LABRAT/modules/salmon_index.nf
@@ -6,7 +6,7 @@ def options    = initOptions(params.options)
 
 process SALMON_INDEX {
         publishDir "${params.outdir}/labrat/salmon", mode: params.publish_dir_mode
-        container "quay.io/biocontainers/salmon:0.14.0--h86b0361_1"
+        container "quay.io/biocontainers/salmon:1.6.0--h84f40af_0"
 
         input:
         path txfasta
@@ -16,6 +16,6 @@ process SALMON_INDEX {
 
         script:
         """
-        salmon index -t $txfasta -i txfasta.idx --type quasi -k 31 --keepDuplicates
+        salmon index -t $txfasta -i txfasta.idx -k 31 --keepDuplicates
         """
 }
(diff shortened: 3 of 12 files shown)


execution_workflows/LABRAT/modules/salmon_quant.nf
```diff
@@ -7,7 +7,9 @@ def options    = initOptions(params.options)
 process SALMON_QUANT {
         tag "$sample"
         publishDir "${params.outdir}/labrat/salmon/", mode: params.publish_dir_mode
-        container "quay.io/biocontainers/salmon:0.14.0--h86b0361_1"
+        container "quay.io/biocontainers/salmon:1.6.0--h84f40af_0"
```
yuukiiwa, line 10: `LABRAT` mentions it depends on [`salmon 0.14.0`](https://github.com/TaliaferroLab/LABRAT#option-2-manual-installation). I personally think it is good to update `salmon` to the most recent version given that everything runs fine downstream

yuukiiwa (commented):
Thank you so much for adding the extension, @dominikburri! This looks really good! I have two minor comments.

execution_workflows/LABRAT/modules/salmon_index.nf
```diff
@@ -16,6 +16,6 @@ process SALMON_INDEX {
 
         script:
         """
-        salmon index -t $txfasta -i txfasta.idx --type quasi -k 31 --keepDuplicates
+        salmon index -t $txfasta -i txfasta.idx -k 31 --keepDuplicates
```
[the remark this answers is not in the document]
dominikburri: The new version of Salmon does not have this parameter anymore: https://salmon.readthedocs.io/en/latest/salmon.html 
First I retained it but I got an error, then I found that this parameter is not used anymore.

execution_workflows/LABRAT/modules/salmon_quant.nf
dominikburri: At least for the samples I was using it was fine. And LABRAT uses the salmon output, this should not change much. 
However, more testing might be needed to double-check.

dominikburri:
> Thank you so much for adding the extension, @dominikburri! This looks really good! I have two minor comments.

Hi @yuukiiwa , thanks for the comments. I think some things are still missing and it would be great if you could continue on this :) 

Do you think you can take care of the other points in the checklist? Especially the hardcoding of file names? 
And we discussed about the quantification bits, that LABRAT likely does not provide us with quantification of polyA sites but rather the differential usage. Could you reflect that in the README and the output files generated? 

That would be awesome :)

dominikburri:
Do we have more comments on these extensions? Or do you want to incorporate more changes? 
If not, I would be happy to mark the PR as ready and merge it.

yuukiiwa:
Hi @dominikburri, I have addressed all the points above and updated the workflow with `--run_quantification` and `--run_differential` flags. I have tested these flags on three situations: quantification only (`--run_quantification` only), differential analysis only (`--run_differential` only), and both quantification and differential analysis(`--run_quantification` and `--run_differential`). Here is the screenshot:
Screenshot from 2022-01-09 19-15-46

dominikburri:
Hi @yuukiiwa , thanks for your work :) 
I noticed that the test uses files which are not readily available, so I wanted to use those in `tests/test_data`, but it appears a file is missing there (see related issue).

dominikburri:
Hi @yuukiiwa, thanks for pushing this workflow!

I now tested the workflow with the test data in `tests/test_data`. 
- I created *siControl* and *siSrsf3* `fastq.gz` files from the corresponding BAM files with `samtools bam2fq` (as described in the directory's readme).
- I used `gencode_2genes_Chr_prefix.vM18.annotation.gff3` as the genome annotation.
- I also needed to create a fasta file, in practise this is no problem but we don't have one in the test data. So, I just downloaded *chr6* and *chr16* from [NCBI](https://hgdownload.soe.ucsc.edu/goldenPath/mm39/chromosomes/) and concatenated them. The zipped file is 76 MB heavy, so I guess there is no point in uploading it. Also, I downloaded fasta from ncbi but the annotation is from gencode, so I would probably need to replace the fasta anyway (though I don't think it causes problems for this test).

All this I put into `samplesheet_single.csv`.

Running the code below gives me
```bash
nextflow main.nf --input more_samplesheets/samplesheet_single.csv \
    -profile singularity \
    -w work \
    -resume \
    --run_quantification --run_differential \
    --conditionA siControl --conditionB siSrsf3
``` 
for `siControl_siSrsf3_labrat_differential.tsv`:
``` tsv
geneID	pval
ENSMUSG00000045962.16	2.28e-33
ENSMUSG00000022865.14	NA
``` 
Which looks awesome! Though I think it should not include the column names (see [here](https://github.com/iRNA-COSI/APAeval/blob/main/execution_workflows/execution_output_specification.md#format-03)).
Not sure if we run into problems later on if we keep them, so probably best to remove column headers. Maybe @SamBryce-Smith has an opinion on this.  

So should we include the test data in the corresponding directory or should we keep it as is? 
Anyway, it is a cool workflow and I'd be glad to include it and to get 2 reviews.

SamBryce-Smith (commented):
Apologies for the delay on this - will list my main points below as they don't necessarily relate to changes in the code (which looks great by the way - making the .db a target will make a big difference!).

## Quantification BED files

This has been raised before and just want to add my voice - I don't think that LABRAT qualifies for the quantification challenge **given the output files it generates**. Salmon outputs the TPM **per transcript**, and **LABRAT does not output a per polyA site TPM** by aggregating these transcript TPMs. Also want to note that BED file is currently invalid according to our specs as it stands (includes header, NA values in start / end fields).

image

Though it would be possible to generate this ourselves, it is not necessarily straightforward (transcript-to-polyA site assignment is provided in the `numberofposfactors.txt` file, but we'd have to go back to the input GTF file to get the polyA site coordinates). I think this is out of the scope of our execution workflows, so I'm in favour of dropping the quantification BEDs entirely (with a note in the README explaining why).

## Differential output

Completely agree with @dominikburri the header should be removed (other than that looks perfect :))

## Test data

I agree with @dominikburri, I think the test data is probably too heavy to include with the repository. I think it would be helpful to include the steps @dominikburri outlined very clearly above in the main README though. For all of our other workflows we provide instructions for running with test data, so for the sake of consistency I think we should do the same here. Even just bullet points like below are probably sufficient:

1. Generate (single-end FASTQ file from test data BAMs (link test data README)
2. Download mouse genome FASTA (and subset to chr6,chr16 with e.g. Dominik's instructions)
3. Update sample sheet (using `gencode_2genes_Chr_prefix.vM18.annotation.gff3` as annotation file)

(The above is just the broad strokes just to get an idea of what I mean, I'm sure it needs more detail & could be written more clearly).

## Specify GFF3 & FASTA in config not sample sheet

Again this is a request in the interests of consistency with other workflows. I'd prefer the input GFF3 & FASTA to be specified once in the main `nextflow.config` rather than once per row in the sample sheet. I can't ever envisage wanting to use different annotations for samples you want to directly compare to one another (i.e. in a differential usage analysis). Having to duplicate the same path multiple times is a little annoying and comes with a slim chance of making typos/minor errors.


Minor point - LABRAT has now been published, would be good to update the citation at the top of the README to this vs the pre-print (not essential though).

https://bmcgenomics.biomedcentral.com/articles/10.1186/s12864-021-07781-1

Thanks again for both of your efforts @dominikburri @yuukiiwa, looking super good overall!

execution_workflows/LABRAT/modules/labrat_maketffasta.nf
```diff
@@ -8,11 +8,14 @@ process LABRAT_MAKETFFASTA {
         publishDir "${params.outdir}/labrat", mode: params.publish_dir_mode
         container "quay.io/biocontainers/labrat:0.2.2--pyhdfd78af_0"
 
+        //time '8h'
```
faricazjj, line 11: does this comment mean the time taken to run this particular process is 8h or what does this comment mean?

execution_workflows/LABRAT/modules/salmon_quant.nf
```diff
@@ -7,7 +7,9 @@ def options    = initOptions(params.options)
...
+
+        cpus 8
```
faricazjj, line 12: I'm not familiar with salmon but is 8 cpus the standard for this process or is it a good idea to parameterize this in the config file so the user can adjust the number if they don't want to use the default 8 cpus for this process?

execution_workflows/LABRAT/modules/labrat_calculatepsi.nf
```diff
@@ -8,20 +8,26 @@ process LABRAT_CALCULATEPSI {
         publishDir "${params.outdir}/labrat", mode: params.publish_dir_mode
         container "quay.io/biocontainers/labrat:0.2.2--pyhdfd78af_0"
 
+        cpus 3
```
faricazjj, line 11: same question here about hardcoding the numebr of cpus for this process as the question below :D

faricazjj (commented):
Thanks @dominikburri this looks great! I just have a couple clarification questions :)

execution_workflows/LABRAT/modules/labrat_calculatepsi.nf
```diff
@@ -8,20 +8,26 @@ process LABRAT_CALCULATEPSI {
...
+        memory '16 GB'
+
```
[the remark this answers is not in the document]
dominikburri: Hm yeah I probably set this because I believed the task could not profit from more. 
This was on some real data, so should've worked with that. 
I was not aware that the memory and cpu are treated as actual values and not as limits as in snakemake.
Of course I can remove the statement, shall I do this?

execution_workflows/LABRAT/modules/labrat_calculatepsi.nf
dominikburri: Yeah I thought it made sense when executing some samples. 
Though as mentioned below, I can remove this statement :)

execution_workflows/LABRAT/modules/labrat_maketffasta.nf
dominikburri: yes, this process takes more than the default 4h, so it would fail. @yuukiiwa commented it out. 
I can remove it altogether.

execution_workflows/LABRAT/modules/salmon_quant.nf
dominikburri: My understanding is, that this number is treated as the maximum. If you supply less cpus somewhere in the config or on the CLI, it would only use those. 
So I think salmon does not profit from more than 8 cpus, however this is subjective. 
Again, I can remove this statement. The one below with `${task.cpus}` should still work though.

dominikburri:
I removed the hardcoded cpu and memory limits. I replaced it with process labels from `base.config`. It appears this is from  [nf-core rnaseq](https://nf-co.re/rnaseq/usage#resource-requests). 
I included for *maketffasta* the long one, which bumps the execution time from 4h to 20h. Because this process really takes longer with the full samples. 
For *calculatepsi* I include `process_low`, as it comes with 2 cpus. This process is using multiple cores, I checked with the report. Seems to be hardcoded in LABRAT, so to ease it, I include this label. 

These labels take the maximum possible cpu and memory limits, so either from the process or the system maximum defined in `nextflow.config`.
Now this is pretty much the same as it is in Snakemake, it just requires more manual work :)

SamBryce-Smith (approved):
Updates look great, tested the different workflows according to instructions and everything worked smoothly!

I have a minor quibble about the `--run_quantification` flag still producing the quantification BED file even though we know it's currently incompatible with the challenge, but I don't consider this a big enough deal to block the merge. After all, the README makes it very clear it is not compatible so the file can be ignored regardless.

dominikburri:
> Updates look great, tested the different workflows according to instructions and everything worked smoothly!
> 
> I have a minor quibble about the `--run_quantification` flag still producing the quantification BED file even though we know it's currently incompatible with the challenge, but I don't consider this a big enough deal to block the merge. After all, the README makes it very clear it is not compatible so the file can be ignored regardless.

Thanks for the quick review! 
Yes, I see your point and I know it's not the prettiest solution, but I did not want to fiddle with it too much, as @yuukiiwa was adding this and we could use it in the future.

faricazjj (approved):
LGTM @dominikburri thank you! :D

PR #202, opened 2021-12-30
Refactor APAtrap to include mandatory run modes

### LINKED ISSUE #201: Refactor APAtrap execution workflow to include run modes in the config file
APAtrap qualifies for identification, quantification, and differential challenges. Right now, it only has a run_differential boolean in its modules.config file to indicate whether to run identification and quantification when set to false, or run differential when set to true. 
Accroding to issue #189, we would like to have run_identification, run_quantification, and run_differential booleans to obtain  any combination of output files. This allows for more flexibility for the user.

faricazjj:
Fixes #201

## Checklist

- [x] I have tested my code


execution_workflows/APAtrap/subworkflows/run_differential.nf
```diff
@@ -0,0 +1,60 @@
...
+
+    ch_predict_apa_input
+        .first()
+        .set{ ch_predict_apa_input }
+
+    /*
+        Perform second step of APAtrap: infer all potential APA sites
+    */
+    PREDICT_APA( ch_sample_bedgraph_files_dir,
+                ch_predict_apa_input )
```
dominikburri, line 43: This seems to be the main part of identification and quantification. 
Does this mean that when running only differential it would run ident. and quant.? 
And would I be able to run differential from already present results? Based on the test data and setting all 3 to `true`, identification & quantification and differential run separately.

I guess we would want to run benchmarks from the same input files anyway, correct?

dominikburri (commented):
Thank you @faricazjj for the great work!
I tested the code based on the readme and it worked like a charm :) I set the modes `run_differential=true`, `run_identification=true` and all 3 to true, and each worked. 
I think it's good to go, but I don't remember whether differential usage should be able to use own input files rather than run from the BAM files. My comment is related to this.

execution_workflows/APAtrap/subworkflows/run_differential.nf
faricazjj: Good point! So running differential would require identification and quantification to run too but in a different command compared to when we're running identification and/or quantificiation only. This is because when running differential, we want samples of the same condition to be grouped together for indentification and quantification. On the other hand, without running differential, we want each sample to be ran individually no matter the condition. The difference in the commands are found inside the module (e.g. https://github.com/iRNA-COSI/APAeval/blob/main/execution_workflows/APAtrap/modules/predict_apa.nf). This difference in commands is why when we set all three settings to be true, we have to re-run identification and quantification when running differential instead of runinng differential from the already present results

dominikburri (approved):
Thanks for the clarifications, I think it is good to go.

yuukiiwa (approved):
Everything looks really good, @faricazjj!! :)

PR #203, opened 2021-12-31
feat: Isoscm execution workflow

### LINKED ISSUE #34: Execution Workflow: IsoSCM
## WHAT    
**Write** execution workflow for [IsoSCM](https://github.com/shenkers/isoscm). Use the provided [small files for testing](https://github.com/iRNA-COSI/APAeval/tree/main/tests/test_data) (running the workflow on real data is a different issue).

## CHECKLIST   
* [ ] Comment your code

* [ ] Output: Adhere to output specification for Identification challenge
Fields:

* [ ] Does IsoSCM perform PAS quantification? If yes:

    > - gene ID

faricazjj:
Fixes #34

This PR the completion of this [PR](https://github.com/iRNA-COSI/APAeval/pull/157)

**Workflow with star genome generate and star alignment:**
Screen Shot 2022-03-30 at 11 34 08 AM

**Workflow without star genome generate but with star alignment:**
Screen Shot 2022-03-30 at 11 06 50 AM

## Checklist

- [x] I have tested my code


execution_workflows/IsoSCM/modules/star_alignment.nf
```diff
@@ -0,0 +1,44 @@
...
+    output:
+    tuple val(sample), val(strand), path(star_out_bam), path(star_out_bai), emit: ch_aligned_bam_files
+
+    script:
+    fastq_file = sample + ".fastq"
+    star_out_bam = sample + ".Aligned.sortedByCoord.out.bam"
+    star_out_bai = sample + ".Aligned.sortedByCoord.out.bam.bai"
+    """
+    samtools bam2fq $bam > $fastq_file
+
```
SamBryce-Smith, line 31: 1- I think this should be a separate process. If have option to set as temporary (i.e. delete after alignment complete) that would be ideal to save disk space. I think should also suffix with .fastq.gz so files are automatically gzipped (this happens with `samtools fastq`, not sure about `samtools bam2fq` command isn't in latest documentation)

2- Have to be very careful here I think with how we pull from the BAM too. As it stands, all reads will be pulled to a single FASTQ file. This is fine for single-end alignments, but for paired-end alignments you will lose the paired information and align all reads individually (STAR asks you to pass paired-end alignments consecutively on the command line). I suspect if we want to extract paired-end reads properly we will also have to sort the input BAM file by name first (e.g. as for [`samtools fastq`](http://www.htslib.org/doc/samtools-fasta.html)). Would need to add option to config asking user to specify whether input alignments are single-end / paired-end too

image

execution_workflows/IsoSCM/modules/star_alignment.nf
```diff
@@ -0,0 +1,44 @@
...
+    STAR \
+      --runThreadN $task.cpus \
+      --outSAMtype BAM SortedByCoordinate \
+      --genomeDir $star_index_file \
+      --outSAMstrandField intronMotif \
+      --readFilesIn $fastq_file \
+      --readFilesCommand cat \
+      --outFileNamePrefix bam_files/$sample. 
+
+    cat bam_files/$star_out_bam > $star_out_bam
```
SamBryce-Smith, line 41: not sure why this cat step is needed?

SamBryce-Smith (commented):
Thanks for this @faricazjj, great stuff!

I haven't been able to get a test run going yet, mainly because I'll need to run the indexing step on our HPC so this review is incomplete. Some general comments below:

- I don't think the README mentions at any point the reason why you might need to realign with STAR. I'm aware from previous PR that this is because the ['XS strand tag' is required by IsoSCM](https://github.com/shenkers/isoscm/issues/9)
- Is it possible to run the workflow **without realignment**? This would be for cases where the users know that their BAM files contain this tag (hopefully we can do this for APAeval datasets to avoid the extra computation). 
    - It would be very nice (but not required) to have a pre-processing script that checks BAM files for these flags in this case, but that wouldn't be straightforward so I think we can be **very explicit on the input file requirements** and trust the user to use the correct settings (linking to GH issue for specific error that arises if not).

I have a few more specific comments outlined in other comments :)

execution_workflows/IsoSCM/modules/star_genome_generate.nf
```diff
@@ -0,0 +1,39 @@
...
+    """
+    [ -d $star_index_dir ] || mkdir $star_index_dir
+
+    STAR \
+      --runMode genomeGenerate \
+      --runThreadN $task.cpus \
+      --genomeDir $star_index_dir \
+      --genomeFastaFiles $fasta_genome_file \
+      --sjdbGTFfile $gtf_genome_file \
+      --sjdbOverhang 74
```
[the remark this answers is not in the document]
faricazjj: oh yes good point! :D

execution_workflows/IsoSCM/modules/star_alignment.nf
faricazjj: ahh added it to see what the output bam file looked like but not needed in the actual workflow, removed! :D

execution_workflows/IsoSCM/modules/star_alignment.nf
faricazjj: Thanks for pointing this out @SamBryce-Smith, I wasn't familiar with single and paired end reads and this was so helpful! :D I've updated the workflow to include single/paired information in the sample sheet so that each sample runs two different fastq and alignment steps based on the information

execution_workflows/IsoSCM/README.md
```diff
@@ -0,0 +1,71 @@
...
+(e.g. to ensure docker is running, run the command `docker --help` and a help message should be printed).
+Additionally, make sure that line 49 in IsoSCM/nextflow.config file `docker.enabled=true` is uncommented while line
+52 `singularity.enabled=true` is commented out
+
+## Singularity
+To run with singularity, comment out line 49 in IsoSCM/nextflow.config file `docker.enabled=true` and make sure that line
+52 `singularity.enabled=true` is uncommented
+
+### Parameters
+Parameters used to run the two steps of DaPars are specified in conf/modules.config file.
```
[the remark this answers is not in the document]
faricazjj: ah yes! updated!

execution_workflows/IsoSCM/README.md
```diff
@@ -0,0 +1,71 @@
...
+The [application download](https://github.com/shenkers/isoscm/releases/tag/IsoSCM-2.0.12) is free to download
+The [documentation](https://github.com/shenkers/isoscm) was used as a reference to 
+create the nextflow pipeline
+
+## Running IsoSCM workflow
+
+
+### Download genome fasta file
+IsoSCM workflow runs STAR genome generate for STAR alignment. This step requires a genome fasta file. To run with test data, run the following to download GRCm38 genome fasta file:
+`wget https://ftp.ebi.ac.uk/pub/databases/gencode/Gencode_mouse/release_M18/GRCm38.p6.genome.fa.gz`
```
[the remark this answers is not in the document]
faricazjj: Good point!

faricazjj:
> Thanks for this @faricazjj, great stuff!
> 
> I haven't been able to get a test run going yet, mainly because I'll need to run the indexing step on our HPC so this review is incomplete. Some general comments below:
> 
> * I don't think the README mentions at any point the reason why you might need to realign with STAR. I'm aware from previous PR that this is because the ['XS strand tag' is required by IsoSCM](https://github.com/shenkers/isoscm/issues/9)
> * Is it possible to run the workflow **without realignment**? This would be for cases where the users know that their BAM files contain this tag (hopefully we can do this for APAeval datasets to avoid the extra computation).
>   
>   * It would be very nice (but not required) to have a pre-processing script that checks BAM files for these flags in this case, but that wouldn't be straightforward so I think we can be **very explicit on the input file requirements** and trust the user to use the correct settings (linking to GH issue for specific error that arises if not).
> 
> I have a few more specific comments outlined in other comments :)

Thank you so much for the detailed review and suggestions @SamBryce-Smith ! :D 
I've updated the PR with your suggestions and can be summarized into these points:

> 1. Star alignment step used to be mandatory but is now optional, the user has the option to set a parameter in modules.config file to true or false. If set to true, star alignment will run.
> 2. Depending on whether the sample contains paired/single end reads, the sample will be treated differently when converting bam to fastq and when running star alignment
> 3. The step to convert bam to fastq is a separate step from star alignment
> 4. The step to get bai file is a separate step from star alignment
> 5. Updated README to address the XS tag requirement for isoscm and added descriptions for the newly added parameters

execution_workflows/IsoSCM/modules/bam_to_fastq.nf
```diff
@@ -0,0 +1,37 @@
...
+                            -fq $fastq1 \
+                            -fq2 $fastq2
+    	"""
+    }
+    else {
+        """
+        bedtools bamtofastq -i $bam -fq $fastq1
+        cp $fastq1 $fastq2
+        """
+    }
```
SamBryce-Smith, line 36: Is the `cp $fastq1 $fastq2` needed for pipeline plumbing? If so happy for it to stay, otherwise it feels a bit unnecessary to  duplicate the file (which could be in the GB range) if STAR doesn't use it? If for pipeline plumbing (i.e. nextflow needs to know file has been generated) then maybe could just `touch $fastq2` to save space?

Also, do we need the fastqs to be uncompressed? Again it's not a critical change, but would save on disk space requirements for runs on full datasets. For STAR you can also provide the `--readFilesCommand zcat` or `--readFilesCommand gunzip -c` so it knows to decompress the input files.

SamBryce-Smith (commented):
Just submitting a few minor comments know whilst I try and get a test run working!

faricazjj (commented):
Thanks for another round of review @SamBryce-Smith ! :D I've fixed the typos and addressed the comments on bam_to_fastq and star_alignment processed :)

dominikburri:
Hi @faricazjj, I try to run the test data and downloaded `GRCm38.p6.genome.fa` as it is described in the readme and copied it into `tests/test_data/` and set `run_star_genome_generate = true` in `modules.config`. I adjusted `path_to` occurrences, as described. I also adjusted max memory and cpus in `nextflow.config` such that I can run it on my computer (10 cores, 30GB).

However, running it gives me this error:
```log
Error executing process > 'EXECUTE_ISOSCM:PREPROCESS_FILES:STAR_GENOME_GENERATE (1)'
Caused by:
  Process `EXECUTE_ISOSCM:PREPROCESS_FILES:STAR_GENOME_GENERATE (1)` terminated with an error exit status (104)

Command executed:

  [ -d star_index ] || mkdir star_index  
  STAR       --runMode genomeGenerate       --runThreadN 10       --genomeDir star_index       --genomeFastaFiles GRCm38.p6.genome.fa       --sjdbGTFfile null       --sjdbOverhang 100

Command exit status:
  104

Command output:
  FATAL STAR --runMode genomeGenerate --runThreadN 10 --genomeDir star_index --genomeFastaFiles GRCm38.p6.genome.fa --sjdbGTFfile null --sjdbOverhang 100
  Mar 31STAR version: 2.7.10a_alpha_220314   compiled: 2022-03-30T13:47:31+00:00 :/STAR/source
  Mar 31 14:24:25 ..... started STAR run
  Mar 31 14:24:25 ... starting to generate Genome files  Mar 31 14:24:56 ..... processing annotations GTF

Command error:  
  FATAL error, could not open file pGe.sjdbGTFfile=null
  
  Mar 31 14:24:56 ...... FATAL ERROR, exiting
```
Can you help me with that? Did I properly set up the test data? Did you encounter the same issue?

dominikburri (approved):
Hi @faricazjj, thank you for the fix!

The test samples now run, with generating the STAR index (first run) and without (second run). 
I first ran it with Docker and then with Singularity, both working without problems.

faricazjj:
> Hi @faricazjj, thank you for the fix!
> 
> The test samples now run, with generating the STAR index (first run) and without (second run). I first ran it with Docker and then with Singularity, both working without problems.

Thank you so much for reviewing and approving! @dominikburri 👯

execution_workflows/IsoSCM/bin/postprocess_identification.py
```diff
@@ -0,0 +1,46 @@
...
+            chrom = row[0]
+            start = row[3]
+            end = int(start) + 1
+            strand = row[6]
+            name = '|'.join([chrom, str(start)+":"+str(end), strand])
+            score = "."
+            output = (chrom, start, str(end), name, score, strand)
+            identification_out.write("\t".join(output) + "\n")
+
+    identification_out.close()
```
SamBryce-Smith, line 37: Mr. One-off is back again :stuck_out_tongue:

The input file appears to be in GTF format (e.g. `results/isoscm/isoscm_assemble_outdir/control_r1/identification_out` & checking `modules/postprocess_identification.nf`):
image

GTF & BED files follow different conventions for reporting coordinates. GTF starts at 1, start and end are included included in range whilst BED starts at 0, start is included but end is excluded from the range. To convert GTF/GFF -> BED, you just need to subtract 1 from the start coordinate. 

P.S. I always like to refer to this [cheat sheet](https://gist.github.com/ilevantis/6d6ecf8718a5803acff736c2dffc933e#gff-vs-bed-indexing) for things like this as it has a nice visualisation (and I have to sanity check these things every time).

execution_workflows/IsoSCM/README.md
```diff
@@ -0,0 +1,86 @@
...
+- Run the pilot benchmark nextflow pipeline with nextflow main.nf --input samplesheet_example_files.csv
+
+## Output & post-processing
+By default, each IsoSCM run results in an identification challenge file located under IsoSCM/results/isoscm/challenges_outputs/[sample]_identification_output.bed
+
+## Notes
+- Running STAR genome generate requires 30-40GB of memory
+- Running STAR genome generate takes 20-25 minutes and is only required once per genome
+- IsoSCM doesn't qualify for quantification challenge because IsoSCM assemble step outputs mean read density in the segments upstream and downstream of the polyA site but doesn't output the TPM value for each identified site that is required for quantification challengenge output  
+- IsoSCM doesn't qualify for differential challnege because the compare step provides a confidence score representing the likelihood that a change-point occurs at the given position, compared to a null model that no change-point at this position. However, this doesn't exactly equal to the significance of differential polyadenylation site usage, which is needed for differential challenge output
```
SamBryce-Smith, line 83: I fully agree this means it does not represent significance of differential pA usage. The closest thing to a change in usage metric is the 'differential_usage' column in the output file, but as the README states that is just an effect size difference and no statistical test is performed on the significance of that change in usage (which is what we want to assess with the differential usage challenge). 

Also seems like `compare` can only handle comparisons between 2 samples (i.e. it cannot handle replicates), which is another factor that makes it difficult to adapt to our datasets (author suggests merging all replicates into a single BAM file https://github.com/shenkers/isoscm/issues/8). I believe lack of ability to handle replicates for differential usage may have been one of the reasons we haven't pressed ahead with MISO).

Apologies for a bit of a ramble, feel like I'm not really suggesting anything. Maybe just to clarify that isoSCM doesn't perform any statistical tests on the significance of the difference in polyA site usage. Feel free to ignore this haha

SamBryce-Smith (commented):
Thanks @faricazjj! Super close to being approved on my end :)

First things first, docs are super clear and all of my test runs were successful!

Successful test run with generating the STAR index (Docker):
image

Also had another successful test run skipping the index step (Docker), and tried Singularity for the full pipeline too and it worked perfectly!

I just have a quick comment on extracting the polyA sites from the output GTF. If you agree it should be a very quick fix so I suspect we can turn around the reviews pretty quickly once you find the time!

execution_workflows/IsoSCM/bin/postprocess_identification.py
faricazjj: hahahahaha ahhh good catch! thank you so much @SamBryce-Smith!! :D also thank you for sharing the cheat sheet!

execution_workflows/IsoSCM/README.md
faricazjj: Thank you for spilling your thoughts on this! I took some time thinking about this trying to see if we can include isoscm in differential challenge by using any of the outputs they provide so it's nice to have an extra confirmation from you! And yes you're right compare can only handle comparisons between 2 samples and no replicates so it would be difficult to use with our datasets. It might be doable by adding an extra process in the workflow but since it doesn't output significance that we need anyway so phewh not a worry for now :p 

But yes in summary what I've found is isoscm doesn't perform any statistical test on the significance of the difference in polyA site usage and I confirmed this with the author too :)

execution_workflows/IsoSCM/README.md
```diff
@@ -0,0 +1,86 @@
...
+IsoSCM is compatible with BAM files generated by many short read mappers but requires that spliced reads include the XS flag specifying the inferred strandedness of the spliced read. 
+
+As such, if the BAM input files provided do not include XS flag, this workflow provides the option to run STAR alignment to output BAM files compatible with IsoSCM. The STAR alignment is split into two steps: 
+1. STAR genome generate
+2. STAR alignment
+
+The STAR genome generate step is required to run only once per genome and takes around 25 minutes to complete. This outputs a star genome index folder. This step can be skipped by setting the parameter run_star_genome_generate in conf/modules.config to false if the user already has a star genome index folder they would like to be used for STAR alignment step. The star genome index folder provided by the user is specified by the absolute path in star_genome_index parameter.
+
+The STAR alignment step can successfully run if the user runs STAR genome generate step as part of the workflow or provides a valid star genome index folder to be used. The user also has the option to skip this step by setting run_star_alignment parameter to false.
```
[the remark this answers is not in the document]
faricazjj: Thank you for digging the issue reported on isoscm regarding the error when there's no xs tag, i'll add it to the readme :)

faricazjj:
@dominikburri @SamBryce-Smith thank you guys for the last round of review! I updated the readme and made a minor change that Sam pointed out in postprocess_identification.py file's lines 30, 31, and 35 and no other changes aside from these. We're so close to being done! 😆

dominikburri (approved):
Thank you so much @faricazjj!

SamBryce-Smith (approved):
Great stuff @faricazjj thanks so much for the speedy changes! Good to go on my end :smile:

PR #206, opened 2022-01-05
docs: Update execution workflow readme method execution

### LINKED ISSUE #189: docs: Update readme to include mandatory parameters for execution workflow
Each execution workflow should have parameters on which benchmark to run in the config file. The parameters are:
run_identification = True/False
run_quantification = True/False
run_differential = True/False

Only parameters for benchmarks that the workflow qualifies for should be included. For example, if a workflow only qualifies for identification and differential, it should only include:
run_identification = True/False
run_differential = True/False

This allows for the workflow to output any combination of benchmarks that it qualifies for, as specified by the user.

Readmes to be updated:
Execution workflow readme
Nextflow template readme
Execution workflow dsl1 template readme
Execution workflow dsl2 template readme

faricazjj:
Fixes #189 

## Checklist

- [x] I have tested my code


dominikburri (approved):
Thank you! Should be fine I think, also since we include the changes in some workflows already :)

SamBryce-Smith (approved):
Thanks for this @faricazjj, agree with Dominik good to add a note here then have more specific detail in the template directories.

faricazjj:
@dominikburri @SamBryce-Smith  thanks for reviewing you guys! :D

PR #208, opened 2022-01-05
Bug fix: Dapars parsing output with wrong cast to string

### LINKED ISSUE #207: Parsing output with wrong cast to string
https://github.com/iRNA-COSI/APAeval/blob/e581d3ffc844961edbc2efd701cc440c4b361709/execution_workflows/DaPars/bin/convert_output.py#L38

The line above leads to an erroneous return statement in this line: https://github.com/iRNA-COSI/APAeval/blob/e581d3ffc844961edbc2efd701cc440c4b361709/execution_workflows/DaPars/bin/convert_output.py#L45

The problem is the cast of the value to `str`, which results in a wrong evaluation of `min`.

Take the example below:
```
sign = [str(1.0), str(0.499), str(0.144), str(7.136e-21)]
min(sign)
```
will give '0.144' instead of the minimum of `7.136e-21`

This can be avoided by casting to **float** (or no casting):
```
sign = [float(1.0), float(0.499), float(0.144), float(7.136e-21)]
min(sign)
```
will give the correct result `7.136e-21`

faricazjj:
Fixes #207 

## Checklist

- [x] I have tested my code

### DIFF
--- a/execution_workflows/DaPars/bin/convert_output.py
+++ b/execution_workflows/DaPars/bin/convert_output.py
@@ -35,14 +35,14 @@ def convert_to_differential(file_in, file_out):
         # e.g. ENSMUST00000203335.1|ENSMUSG00000045962.16|chr6|-
         name = row['Gene'].split("|")[1]
         # p val obtained from P_val column
-        significance = str(row['P_val'])
+        significance = row['P_val']
         if name not in rows:
             rows[name] = [significance]
         else:
             rows[name].append(significance)
     # only get the smallest p-value for the corresponding gene
     for gene in rows:
-        output = [gene, min(rows[gene])]
+        output = [gene, str(min(rows[gene]))]
         differential_out.write("\t".join(output) + "\n")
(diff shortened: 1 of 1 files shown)


dominikburri (approved):
Thanks! The fix seems to work :)
Tested differential usage on this branch and the output file now carries two genes, which based on `dapars_output_All_Prediction_Results.txt` have both the lowest observed p-value. 

Possibly easier to merge this one before the big refactoring.

faricazjj:
> Thanks! The fix seems to work :) Tested differential usage on this branch and the output file now carries two genes, which based on `dapars_output_All_Prediction_Results.txt` have both the lowest observed p-value.
> 
> Possibly easier to merge this one before the big refactoring.

Thanks for reviewing and approving this PR @dominikburri ! :D Sounds good, I'll aim to merge this fist before the refactoring PR

yuukiiwa (approved):
Thanks, @faricazjj!! This looks good!

PR #210, opened 2022-01-12
Added plot descriptions to benchmark specifications

### LINKED ISSUE #197: [docs] Update benchmark specifications with visualisation requirements
Add information about how metrics should be visualised.
Problems: 

- OpenEBench currently supports 2D scatter plots and bar plots, but it would be useful to have 2D line charts for some metrics
- In some benchmarks there are metrics that can't be plotted together - they probably should be in separate benchmarks

daneckaw:
Fixes #197 

Added plot descriptions and ranking rules to benchmark specifications. Those include 2D line plots that are not yet implemented in OpenEBench as well as alternative plots in case they cannot be implemented on time.

## Checklist

- [x] I have tested my code


summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md
```diff
@@ -52,6 +59,64 @@ Fields:
...
+
+Optional plots:
+
+5. **bar plot** visualizing **AUC** of ROC plot with single AUC value for each tool.
+
+**X axis** - benchmarked tool  
+**Y axis** - AUC
+
+Ranking: The best performing tool is the one with the highest AUC value.
+
```
mzavolan, line 118: Since #5 can be addressed with the current functionality of OpenEBench, I would use this instead of #4

summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md
```diff
@@ -69,13 +134,20 @@ description of each attribute-value pair:
...
   
 ## Metrics
   
   | # | Description | Unit | Compute from | Transformations | Type after transformations | Additional comments |
   | --- | --- | --- | --- | --- | --- | --- |
-  | 1 | Sensitivity | N/A | Output 1 | Read file, parse JSON and extract attribute `sensitivity` that has type `vector` of `float`. The values should be plotted against a vector of distance cutoff values from 0 to 100 nt with 10 nt intervals. | `array` | N/A |
-  | 2 | FDR | N/A | Output 1 | Read file, parse JSON and extract attribute `FDR` that has type `vector` of `float`. The values should be plotted against a vector of distance cutoff values from 0 to 100 nt with 10 nt intervals. | `array` | N/A |
+  | 1 | Sensitivity | N/A | Output 1 | Read file, parse JSON and extract attribute `sensitivity` that has type `vector` of `float`.  | `array` | N/A |
+  | 2 | FDR | N/A | Output 1 | Read file, parse JSON and extract attribute `FDR` that has type `vector` of `float`.  | `array` | N/A |
+  | 3 | FDR | N/A | Output 1 | Read file, parse JSON and extract attribute `FPR` that has type `vector` of `float`. | `array` | N/A |
```
mzavolan, line 148: Should this be FPR?

summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md
```diff
@@ -69,13 +134,20 @@ description of each attribute-value pair:
...
+  | 4 | Precision | N/A | Output 1 | Read file, parse JSON and extract attribute `FDR` that has type `vector` of `float`. | `array` | N/A |
+  | 5 | AUC | N/A | Output 1 | Read file, parse JSON and extract attribute `FDR` that has type `float`. | `vector` | N/A |
```
mzavolan, line 150: The attribute bane seems to be incorrect for both #4 and #5

daneckaw:
The original templates for benchmark specifications included output in json and we never questioned/revised it. It seems like we don't use those files for anything (for now) and I'm not sure if we can get this output from OpenEBench. Should we just remove it or use a different format? It would be useful to have benchmarking results if we need to make any figures not available on OEB, but I think we can just query the OEB database for that? @mzavolan @AsierGonzalez @dominikburri

AsierGonzalez:
Good point, @daneckaw. You are right in that we haven't reviewed the output files' format. Indeed, that is a key aspect of making the summary workflows compatible with OpenEBench since the output files **must** adhere to the OpenEBench data model. I will summarise here what the output file should look like, but if you want to know more about the OpenEBench data model, you can check the [documentation in Read the Docs](https://openebench.readthedocs.io/en/latest/technical_references/1_scientific_data_model.html) or the [GitHub repository](https://github.com/inab/benchmarking-data-model/tree/2.0.x/json-schemas) (there note we're still using version 1.0.0).

In fact, the summary workflow creates no less than three output files, one per step:
workflow_schema_outputs

I will only talk about the "aggregation dataset" (aka "consolidation dataset"), given that the contents of the "participant dataset" and the "assessment dataset" are replicated in it. This file contains a JSON array with three different types of objects:

1. Participant: It contains basic information about the benchmarked tool and the challenge, namely the name of the tool, id of the community and the challenges and the time stamp.
2. Assessment: It contains the value of a metric. There will be one assessment object per challenge, tool and metric.
3. Aggregation: This object summarises the results of all the tools. There is one object per challenge that contains an array with as many items as tools.

An example of such a file can be found [here](https://github.com/inab/TCGA_benchmarking_workflow/blob/1.0.8/TCGA_sample_out/benchmarking_data_model_export/consolidated_result.json).

One of the peculiarities of APAeval is that you want to measure the same metrics for different window sizes. This has never been done before and although the data model does not have any restrictions about this, we will need to see what the best way to do this would be.

AsierGonzalez:
By the way, I forgot to mention that you will be able to access the benchmarking results in one of two ways:

1. Download them directly from the workspace in the [OpenEBench Virtual Research Environment](https://openebench.bsc.es/vre/home/).
2. Query the OpenEBench REST or GraphQL APIs.

1. is much easier but we can discuss in the future which is more appropriate depending on your needs.

daneckaw:
Updates:
- changed example output files (assessment and aggregation) based on examples here: https://github.com/inab/TCGA_benchmarking_dockers/tree/master/sample_results @AsierGonzalez does it make any sense?
- merged I2, I3, I5 into I2 and Q2, Q3 into Q2 because they use the same metrics and the same input format, just different input types (real vs simulated data). This should also clarify which python scripts are used for which metrics.
- clarified the usage of `window` parameter for metrics calculation.
- added a section on how our specifications map to OEB benchmarking events and challenges.

summary_workflows/identification/I2_benchmark/example_files/aggregation_out.json
```diff
@@ -0,0 +1,29 @@
+{
+    "_id": "OEBid",
+    "challenge_ids": [
+        "datasetA_sensitivity_FPR_10nt"
```
AsierGonzalez, line 4: I think this would be only `datasetA`, it must match the `challenge_id` in the `assessment_out.json` file

AsierGonzalez (commented):
Nice work @daneckaw! I have added some small comments to the spec file and the output assessment and aggregation files, nothing major.

The one thing we need to give some thought to is if we really need to define one metric per window size as it [currently appears in the spec](https://github.com/daneckaw/APAeval/blob/plot_specs/summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md#additional-info-1). My feeling is that instead of `Precision_10nt`, `Precision_50nt` and `Precision_100nt` we should just have `Precision` and add the window size somewhere with the results. That is only my feeling though, I need to check what consequences each of the alternatives would have.

summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md
```diff
@@ -47,45 +64,146 @@ Fields:
 - **score** - not used, leave as "."
 - **strand** - defines the strand; either "." (=no strand) or "+" or "-".
 
-#### Format 2
+#### Input 2
 
 This BED file contains single-nucleotide position of poly(A) sites identified from the orthogonal 3'end-seq dataset.  
-Fields are the same as in Format 1.
+Fields are the same as in Input 1.
```
ninsch3000, line 70: Should the PAS databases also be declared as input? Or maybe mention that they need to have the same format as the orthogonal 3'end seq data, or whatever else makes sense here

summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md
```diff
@@ -47,45 +64,146 @@ Fields:
...
 ## Outputs
 
-| # | Format | Link | Example data |
-  | --- | --- | --- | --- |
-  | 1 | JSON | [Specification][spec-json] | [Link][out1] |
-  
-### Additional info outputs
+Calculated metrics are saved in JSON file adhering to OpenEBench schema. 
+Assessment output contains values of calculated metrics for a given input dataset.
+Consolidation output contains summarized data in a format suitable for plotting, e.g. single values for barplot or X,Y value pairs for 2D scatter plot.
```
ninsch3000, line 161: Maybe: "Consolidation output contains summarized data *from all benchmarked tools within one challenge* in a format suitable for plotting..."

summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md
```diff
@@ -47,45 +64,146 @@ Fields:
...
+
 ## Outputs
 
-| # | Format | Link | Example data |
-  | --- | --- | --- | --- |
-  | 1 | JSON | [Specification][spec-json] | [Link][out1] |
-  
-### Additional info outputs
+Calculated metrics are saved in JSON file adhering to OpenEBench schema. 
+Assessment output contains values of calculated metrics for a given input dataset.
```
ninsch3000, line 160: Maybe for emphasis: "Assessment output contains values of calculated metrics for a given input dataset... *for one tool*." or "*for the tool being benchmarked" or "Assessment output is generated for each tool separately and contains..."

ninsch3000:
Hi @daneckaw , could you resolve the comments from Asier and me? If they are not relevant you can just write that/click `resolve conversation` below them. And then let me know in a general comment that the PR is ready to merge.
Thanks a lot :)

ninsch3000 (approved):

dominikburri (approved):
Hi @daneckaw, thank you very much for the great work! 
I think it is quite clear, but I still have some questions.

- Are the `deprecated` directories needed, can we delete them? It would make it a clearer appearance.
- Are the identification and quantification benchmark like this complete or will *I4*: `summary_workflows/identification/I4_benchmark/I4_benchmark_specification.md` and *Q4* (not yet in `main`) be included in these files? Maybe I just got confused by the title. I tried  to match up the specification here and in the excel table.  

Anyway, the changes look great and since my and other comments and questions are not crucial to fix right now, it is good to go from my side.

summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md
```diff
@@ -47,45 +59,128 @@ Fields:
...
+
+2. **2D scatter plot** visualizing **TPR and FPR** of poly(A) site identification. Separate plots should be prepared for different values of distance threshold:
+
+- 10 nt
+- 50 nt
+- 100 nt
+
+**X axis** - FPR(d)  
+**Y axis** - TPR(d)  
+where _d - distance threshold_
```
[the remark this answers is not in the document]
daneckaw: There would be 3 separate plots, each for a different distance threshold, with one point per participant showing FPR on X axis and TPR on Y axis. I don't think 3 points per participant in one plot can be reasonably shown in OEB, but if they can (e.g. same color but different shape), it would be great to decrease the number of plots.

summary_workflows/quantification/Q2_benchmark/Q2_benchmark_specification.md
```diff
@@ -1,43 +1,57 @@
-# Q2 benchmark specification
+# Quantification benchmarks specification
```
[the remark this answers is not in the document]
daneckaw: All quantification specifications will be in the same specification file, as I think we should have one specification for one benchmarking event.

summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md
```diff
@@ -1,41 +1,53 @@
-# I2 benchmark specification
+# Identification benchmarks specification
```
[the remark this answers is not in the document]
daneckaw: Yes, all identification specifications will be here. I just need to describe the plots to make them compatible with OEB.

summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md
```diff
@@ -47,45 +59,128 @@ Fields:
...
+**X axis** - FPR(d)  
+**Y axis** - TPR(d)  
+where _d - distance threshold_
+
+Input datasets:
+
+- RNA-Seq data compared with 3'end sequencing data
+- Simulated RNA-Seq data compared with dataset used for simulation
+
+Ranking: The best performing tool is the one with the highest TPR combined with lowest FPR (top left part of the plot) and the worst performing tool is the one with the lowest TPR combined with highest FPR (bottom left part of the plot. The plot should be divided into diagonal quartiles based on the distance from optimal performance. Alternatively, if the plot is divided into square quartiles, the following ranking order should be applied: top-left, top-right, bottom-left, bottom-right.
```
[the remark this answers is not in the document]
daneckaw: This was intended as 3 separate plots (one per distance/window), so no AUC. My idea was that we cannot visualise a proper ROC in OEB, but we might still want to _look_ at how different participants perform for several representative thresholds instead of just having a single AUC number. If the plot type described in plot 4 was available, plot 2 would be unnecessary.

summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md
daneckaw: Correct, databases are not used (for now).

summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md
daneckaw: I thought that in TCGA example the results for all tools for one dataset are reported in the same assessment file, but I'll double check if that's the case and change it in the next PR if necessary.

summary_workflows/identification/I2_benchmark/I2_benchmark_specification.md
```diff
@@ -95,69 +117,92 @@ Ranking: The best performing tool is the one with the highest sensitivity value.
...
-  
-  N/A
+#### Output 1
+
+The OpenEBench assessment file contains the following attributes:
+
+- **\_id** - follows the format: community:challenge\_metric\_tool
+- **challenge_id** - describing the combination of input dataset with indication whether it is based on real or simulated data and any parameters used for metric calculation; e.g. datasetA\_simulated\_10nt
+- **participant_id** - benchmarked tool
+- **stderr** - standard error, set to zero
```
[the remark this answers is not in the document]
daneckaw: It makes sense, I'll correct it in the next PR.

daneckaw:
There are some things to address, summarized here:
- make JSON descriptions follow the file structure
- should window/distance be included in challenge name
- one assessment file for multiple participants or mutliple assessment files?
- specifications for Q4 and I4 missing from new specification format
- remove deprecated folders

If it's OK, I'd prefer to address them in the next PR. Added an issue: #219

PR #217, opened 2022-02-24
Update Dapars Dockerfile

### LINKED ISSUE #216: Dapars error when running on big human genome samples
Dapars errors out when running big human genome samples due to string to integer casting issue which should have been string to float casting. There was already an existing [PR](https://github.com/ZhengXia/dapars/pull/5) reported in Dapars repo to fix the same issue. Yoseph and I contacted the author to have this fixed. The fix has been added by the author, but the current repo is not released on github. As such, we need to update Dapars Dockerfile to download the latest repo that includes this fix.

The error was reported by @dominikburri below:

```
Error executing process > 'EXECUTE_DAPARS:RUN_DAPARS:DAPARS_MAIN (null)'

Caused by:
  Process `EXECUTE_DAPARS:RUN_DAPARS:DAPARS_MAIN (null)` terminated with an error exit status (1)

Command executed:

  python /dapars/src/DaPars_main.py config

Command exit status:
  1

Command output:
  (empty)

Command error:
  [Thu Nov 18 15:22:14 2021] Start Analysis ...
  [Thu Nov 18 15:22:14 2021] Loading coverage ...
  Traceback (most recent call last):
    File "/dapars/src/DaPars_main.py", line 548, in <module>
      De_Novo_3UTR_Identification_Loading_Target_Wig_for_TCGA_Multiple_Samples_Main(sys.argv)
    File "/dapars/src/DaPars_main.py", line 154, in De_Novo_3UTR_Identification_Loading_Target_Wig_for_TCGA_Multiple_Samples_Main
      All_samples_Target_3UTR_coverages, All_samples_sequencing_depths, UTR_events_dict = Load_Target_Wig_files(All_Sample_files, Annotated_3UTR_file)
    File "/dapars/src/DaPars_main.py", line 507, in Load_Target_Wig_files
      curr_sample_All_chroms_coverage_dict[chrom_name][1].append(int(fields[-1]))
  ValueError: invalid literal for int() with base 10: '1.0045e+06'
```

faricazjj:
Fixes #216 

## Checklist

- [x] I have tested my code

### DIFF
--- a/execution_workflows/DaPars/Dockerfile
+++ b/execution_workflows/DaPars/Dockerfile
@@ -12,7 +12,8 @@ RUN apt-get update && apt-get install -y curl unzip python python-dev python3 py
 # Install DaPars
 RUN wget https://github.com/ZhengXia/dapars/archive/master.zip
 RUN unzip -q "master.zip"
-RUN chmod +x /dapars-master/src/*
+RUN mv dapars-master dapars
+RUN chmod +x /dapars/src/*
 ENV PATH=$PATH:/dapars/src/
 
 # Download tools to convert gtf gene model file to bed
--- a/execution_workflows/DaPars/Dockerfile
+++ b/execution_workflows/DaPars/Dockerfile
@@ -10,9 +10,9 @@ ENV DEBIAN_FRONTEND noninteractive
 RUN apt-get update && apt-get install -y curl unzip python python-dev python3 python3-pip r-base bedtools wget
 
 # Install DaPars
-RUN curl -LJO gitreleases.dev/gh/ZhengXia/dapars/latest/dapars_0.9.1.zip
-RUN unzip -q "dapars_0.9.1.zip"
-RUN chmod +x /dapars/src/*
+RUN wget https://github.com/ZhengXia/dapars/archive/master.zip
+RUN unzip -q "master.zip"
+RUN chmod +x /dapars-master/src/*
 ENV PATH=$PATH:/dapars/src/
 
 # Download tools to convert gtf gene model file to bed
@@ -23,7 +23,7 @@ RUN mv gtfToGenePred /usr/local/bin
 RUN mv genePredToBed /usr/local/bin
 
 # Install python packages
-RUN wget wget https://bootstrap.pypa.io/pip/2.7/get-pip.py
+RUN wget https://bootstrap.pypa.io/pip/2.7/get-pip.py
 RUN pip3 install pandas
 RUN python2.7 get-pip.py
 RUN pip2.7 install numpy
--- a/execution_workflows/DaPars/Dockerfile
+++ b/execution_workflows/DaPars/Dockerfile
@@ -10,9 +10,10 @@ ENV DEBIAN_FRONTEND noninteractive
 RUN apt-get update && apt-get install -y curl unzip python python-dev python3 python3-pip r-base bedtools wget
 
 # Install DaPars
-RUN wget https://github.com/ZhengXia/dapars/archive/master.zip
-RUN unzip -q "master.zip"
-RUN mv dapars-master dapars
+RUN wget https://github.com/ZhengXia/dapars/archive/546bdd4d809b097fd7a946d42b56e0efa0228472.zip
+RUN mv *zip dapars.zip
+RUN unzip -q "dapars.zip"
+RUN mv dapars-* dapars
 RUN chmod +x /dapars/src/*
 ENV PATH=$PATH:/dapars/src/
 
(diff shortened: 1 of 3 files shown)


execution_workflows/DaPars/Dockerfile
```diff
@@ -10,9 +10,9 @@ ENV DEBIAN_FRONTEND noninteractive
 RUN apt-get update && apt-get install -y curl unzip python python-dev python3 python3-pip r-base bedtools wget
 
 # Install DaPars
-RUN curl -LJO gitreleases.dev/gh/ZhengXia/dapars/latest/dapars_0.9.1.zip
-RUN unzip -q "dapars_0.9.1.zip"
-RUN chmod +x /dapars/src/*
+RUN wget https://github.com/ZhengXia/dapars/archive/master.zip
+RUN unzip -q "master.zip"
+RUN chmod +x /dapars-master/src/*
 ENV PATH=$PATH:/dapars/src/
```
dominikburri, line 16: should be `ENV PATH=$PATH:/dapars-master/src/`
right now, it throws an `DaPars_Extract_Anno.py: command not found` error, which is because the path setting here is not correct.

dominikburri (changes requested):
Hi @faricazjj, thank you for providing this fix!

There is one small bug in the dockerfile, I annotated it. 

I tried the fix with a local docker image to circumvent the current issue.
I built the dockerfile with `docker build . -t apaeval/dapars:local` and replaced all 10 occurrences of the dockerhub version with the new local one (`container "apaeval/dapars:local"`).
I ran 4 human samples (challenge code AK) with `docker.enabled=true` and the extraction of 3'UTRs now works. But now I get an error for the main function: 
`python: can't open file '/dapars/src/DaPars_main.py': [Errno 2] No such file or directory`
I think `dapars` also has to be replaced by `dapars-master` in `modules/dapars_main.nf`.

Besides, maybe it is easier to define docker containers in `nextflow.config` after line 49 (docker.enabled) with `process.container ="docker.io/apaeval/dapars:latest"` and remove all occurrences in the scripts in `modules/`.

I think these two fixes should be enough but please run the local tests (preferably with singularity).

execution_workflows/DaPars/Dockerfile
```diff
@@ -10,9 +10,9 @@ ENV DEBIAN_FRONTEND noninteractive
 RUN apt-get update && apt-get install -y curl unzip python python-dev python3 python3-pip r-base bedtools wget
 
 # Install DaPars
-RUN curl -LJO gitreleases.dev/gh/ZhengXia/dapars/latest/dapars_0.9.1.zip
-RUN unzip -q "dapars_0.9.1.zip"
-RUN chmod +x /dapars/src/*
+RUN wget https://github.com/ZhengXia/dapars/archive/master.zip
```
AsierGonzalez, line 13: I don't feel very comfortable downloading the `master` branch since it could change without notice, meaning that it isn't reproducible. I think it's a good idea to follow best practices although It's true that it seems unlikely that master will change in the near future. In fact, DaPars does not seem to be actively developed, I don't know if DaPars2 superseded it. The last commit to the repo was made in July 2021 and it only updated the README whereas the last real change to the code was in November 2015, which fixed the issue you mention. It's unfortunate that the author didn't create a new release after that.

Unless you can convince him to create a new release or tag I would suggest downloading the latest commit [546bdd4](https://github.com/ZhengXia/dapars/commit/546bdd4d809b097fd7a946d42b56e0efa0228472) using the url https://github.com/ZhengXia/dapars/archive/546bdd4d809b097fd7a946d42b56e0efa0228472.zip.

AsierGonzalez (commented):
I'd recommend downloading the [latest commit](https://github.com/ZhengXia/dapars/commit/546bdd4d809b097fd7a946d42b56e0efa0228472) instead of `master`. See my comment.

AsierGonzalez (approved):
Thanks for addressing the comment!

dominikburri (approved):
Thank you for fixing this @faricazjj! 

I used the [docker image](https://hub.docker.com/r/apaeval/dapars) as it is provided (tag: latest), adjusted the paths as noted in the readme and ran the local test successfully.
Then I ran the same human samples as before for identification and differential usage with Docker. This did not work as it could not find the files and I figured out that the mounting of the file system was not complete. This is on my side though since I don't have the samples in the working directory. But we will most likely not use Docker or simply have the samples in the work dir, so I didn't bother to solve this.
So instead I switched to use Singularity with adding
```
singularity.enabled=true
singularity.autoMounts = true
```
in `nexflow.config` and ran the same sample again without problems this time!

So good to go from my side!

Issue #220, opened 2022-03-25
Update swf directory structure


ninsch3000:
After merging #182  the directory structures of `summary_workflows/[benchmarking_event]` have to be updated according to the following format:   
```
summary_workflows
    - quantification/
          - quantification_dockers/
          - specification/
              - example_files/
              - Q_benchmark_specification.md
      - identification/
          - identification_dockers/
          - specification/
              - example_files/
              - I_benchmark_specification.md
```

Thus, *all* metrics, inputs and outputs of one benchmarking event are specified within the same document.
related to #219 and #185   
blocked by #182

PR #222, opened 2022-03-30
Specification updates

### LINKED ISSUE #219: Benchmark specifications corrections
Summary of comments from previous PR after restructuring the specifications:
- make JSON descriptions follow the file structure
- should window/distance be included in challenge name
- one assessment file for multiple participants or multiple assessment files?
- specifications for Q4 and I4 missing
- remove deprecated folders

daneckaw:
Fixes #219

Changes:
- update specifications for quantification metrics
- update directory structure according to #220 
- delete deprecated Q files

## Checklist

- [ ] I have tested my code


summary_workflows/differential_usage/specification/D_benchmark_specification.md
```diff
@@ -0,0 +1,133 @@
...
+The OpenEBench consolidation file contains all the information about the new benchmarking run and specifies visualization types (2D scatter plot, barplot), descriptions of metrics used for X and Y axis and (X,Y) value pairs for each challenge and challenge participants.
+ 
+
+[//]: # (References)
+  
+[in1]: ./example_files/input1.tsv
+[in2]: ./example_files/input2.tsv
+[assessment_out]: ./example_files/assessment_out.json
+[aggregation_out]: ./example_files/aggregation_out.json
+[consolidation_out]: ./example_files/consolidation_out.json
```
ninsch3000, line 130: This file doesn't exist

summary_workflows/identification/specification/I_benchmark_specification.md
```diff
@@ -151,36 +200,45 @@ The OpenEBench assessment file contains the following attributes:
...
+The OpenEBench consolidation file contains all the information about the new benchmarking run and specifies visualization types (2D scatter plot, barplot), descriptions of metrics used for X and Y axis and (X,Y) value pairs for each challenge and challenge participants.
 
 [//]: # (References)
   
 [in1]: ./example_files/input1.bed
 [in2]: ./example_files/input2.bed
+[in3]: ./example_files/input3.gtf
 [assessment_out]: ./example_files/assessment_out.json
+[aggregation_out]: ./example_files/aggregation_out.json
 [consolidation_out]: ./example_files/consolidation_out.json
```
ninsch3000, line 239: This file doesn't exist

summary_workflows/differential_usage/specification/example_files/aggregation_out.json
```diff
@@ -0,0 +1,34 @@
+{
```
ninsch3000, line 1: Can you make this a list of 1 aggregation object? ( I mean just square brackets at beginning and end of file)

summary_workflows/identification/specification/example_files/aggregation_out.json
```diff
@@ -1,7 +1,7 @@
 {
```
ninsch3000, line 1: Can you make this a list of 2 aggregation objects, and add a second object with the plot for the 50nt metrics that are present in the assessment_out.json?

summary_workflows/quantification/specification/Q_benchmark_specification.md
```diff
@@ -125,27 +166,39 @@ The OpenEBench assessment file contains the following attributes:
...
 
 
 [//]: # (References)
  
 [in1]: ./example_files/input1.bed
 [in2]: ./example_files/input2.bed
+[in3]: ./example_files/input3.gtf
 [assessment_out]: ./example_files/assessment_out.json
+[aggregation_out]: ./example_files/aggregation_out.json
+[consolidation_out]: ./example_files/consolidation_out.json
```
ninsch3000, line 201: This file doesn't exist

summary_workflows/quantification/specification/Q_benchmark_specification.md
```diff
@@ -65,32 +92,45 @@ Fields:
...
-Ranking: The best performing tool is the one with the highest correlation value.
+Ranking: The best performing tool is the one with the highest correlation value and lowest total expression of non-matched sites (top-left part of the plot) and the worst performing tool is the one with the lowest correlation combined with the highest total expression of non-matched sites (bottom-right part of the plot). The plot should be divided into diagonal quartiles based on the distance from optimal performance. Alternatively, if the plot is divided into square quartiles, the following ranking order should be applied: top-left, top-right, bottom-left, bottom-right.
+
+2. **Bar plot** visualizing **Mean Squared Error** of relative PAS usage. Separate plots should be prepared for different values of distance threshold:
+
+- 100 nt
+
+**X axis** - benchmarked tool  
+**Y axis** - MSE
+
```
dominikburri, line 128: I think you missed section "Input datasets" for this bar-plot.

summary_workflows/identification/specification/I_benchmark_specification.md
```diff
@@ -16,14 +16,22 @@ Comparison of PAS predicted from simulated RNA-Seq data with dataset used for si
...
+Calculating proportion of identified PAS assigned to different genomic features:
+
+1. Poly(A) sites identified based on RNAseq data using the benchmarked tool
+2. GTF file with genome annotation
+
 ### Metrics:
 
 Based on the input data the following metrics are computed:
 
 1. Sensitivity (TPR, True Positive Rate) = TP/(TP+FN)
```
dominikburri, line 28: I just want to be more verbatim here to not risk misunderstandings:

- TP: number of PAS identified by participant that overlap PAS from ground truth within distance threshold.
- FP: number of PAS identified by participant that do not overlap ground truth PAS within distance threshold.
- FN: number of PAS not identified by participant but occurring in ground truth within distance threshold.
- TN: not measurable.
- distance threshold: padding of ground truth PAS in number of nucleotides in genomic positions. Example: PAS is in range [500,520] and distance threshold 10nt, so the checks for overlaps are within [490,530].
Correct @daneckaw ?

summary_workflows/identification/specification/I_benchmark_specification.md
```diff
@@ -44,7 +52,8 @@ All identification challenges belong to the same benchmarking event.
 | :-- | :--- | :--- | :--- | :----------|
 | 1 | BED | [Specification][spec-bed] | [Link][in1] | BED file with PAS identification by the benchmarked tool |
 | 2 | BED | [Specification][spec-bed] | [Link][in2] | BED file with PAS identification from ground truth dataset |
-
+| 3 | GFF | [Specification][spec-gtf] | [Link][in3] | GTF file with genome annotation including 3' UTRs |
```
dominikburri, line 55: probably meant GTF, not GFF.

summary_workflows/identification/specification/I_benchmark_specification.md
```diff
@@ -63,60 +72,71 @@ Fields:
 
 This BED file contains single-nucleotide position of poly(A) sites identified from the orthogonal 3'end-seq dataset.  
 Fields are the same as in Input 1.
+
+#### Input 3
+
+This GTF file contains reference genome annotation for the RNAseq dataset including untranslated regions.
```
dominikburri, line 78: does it need to contain fields like "untranslated_region", "3UTR" in column 3? 
Or could it be the last exon in a transcript? (normally contains the 3'UTR).

summary_workflows/identification/specification/I_benchmark_specification.md
```diff
@@ -63,60 +72,71 @@ Fields:
...
+
+Ranking: The best performing tool is the one with the highest ercentage of genes with correctly identified number of PAS.
 
-Optional plots:
+Optional plots - can be generated outside of OpenEBench:
 
-4. **2D line plot** visualising **TPR and FPR** in the form of [ROC curve][roc-curve], i.e. TPR and FPR calculated for a range of distance thresholds are plotted together. 
+1. **2D line plot** visualising **TPR and FPR** in the form of [ROC curve][roc-curve], i.e. TPR and FPR calculated for a range of distance thresholds are plotted together. 
 
 **X axis** - FPR(d)
```
dominikburri, line 141: I don't think we can calculate FPR=FP / (FP + TN), as it contains true negatives which we cannot assess. Please correct me if I'm wrong!

summary_workflows/differential_usage/specification/D_benchmark_specification.md
```diff
@@ -0,0 +1,133 @@
...
+### Metrics
+
+Based on the input data the following metrics are computed:
+
+1. Sensitivity (TPR, True Positive Rate) = TP/(TP+FN)
+2. FPR (False Positive Rate) = FP/(FP+TN)
+
+TP - true positives - genes identified by the tool and present in the orthogonal dataset  
+FP - false positives - genes identified by the tool and not present in the orthogonal dataset  
+FN - false negatives - genes not identified by the tool but present in the orthogonal dataset
```
dominikburri, line 33: Thank you for the clear specification! 
TN missing though.

Also, I guess you mean with "genes identified", that these genes are considered significant in the sense of differential PAS usage between two conditions? So TP would be the number of genes with FDR < ALPHA (normally 0.05) identified by the participant and orthogonal ground truth dataset. 
But how do we identify differential PAS usage from ground truth?

summary_workflows/quantification/specification/Q_benchmark_specification.md
```diff
@@ -30,8 +37,26 @@ To compute the metrics, the poly(A) sites identified from the RNAseq data by the
...
+2. Total expression of non-matched sites
+
+The expression values of all predicted sites that were not mapped to any ground truth site are summed together. As the PAS expression levels are provided as TPM, these values are relative to the expression of all identified PAS.
+  
+3. Mean Squared Error (MSE) of relative PAS usage calculated from RNAseq-based PAS quantification and orthogonal 3'end seq data
+
+The poly(A) sites identified by the benchmarked tool are mapped to ground-truth PAS in the orthogonal dataset as in point 1.
+The identified PAS are then assigned to genes based on genome annotation, and relative PAS usage is calculated separately for each gene.
+PAS that do cannot be assigned to any genes are discarded.
+The MSE is then calculated globally for all poly(A) sites assigned to genes.
```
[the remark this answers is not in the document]
mzavolan: In the ground truth there is no restriction on the location of poly(A) sites within a gene.

However, is an MSE metric necessary? Isn't a correlation of PAS usage more fine-grained?

summary_workflows/identification/specification/I_benchmark_specification.md
mzavolan: If we take the ground truth at face value, TN are also measurable. For sure this will be the case for simulation data.

summary_workflows/identification/specification/I_benchmark_specification.md
mzavolan: See above

summary_workflows/differential_usage/specification/D_benchmark_specification.md
mzavolan: Again, I think TN can be determined if the ground truth is really what the name implies.

summary_workflows/identification/specification/I_benchmark_specification.md
dominikburri: @mzavolan, @ninsch3000  and I discussed today about TNs. The conclusion was that if the gene is not expressed in the given data, the polyA site is not identified and is thus a negative. Then a true negative would be if the prediction also does not give a polyA site on that location. 
This pattern would most likely only hold for simulation data, where from a selection of polyA sites reads are generated.

summary_workflows/differential_usage/specification/D_benchmark_specification.md
dominikburri: From the same discussion today. A TN for the differential usage has a different meaning compared to the one in the identification benchmark.

summary_workflows/identification/specification/I_benchmark_specification.md
daneckaw: I was thinking about it and for most cases TN would not be possible to measure, so in this version I removed the FPR metric that uses TN. Somehow I forgot to update the optional ROC curve and change it to Precision-Recall curve that does not need TN.

summary_workflows/identification/specification/I_benchmark_specification.md
daneckaw: No, it just needs something that corresponds to 3'UTR. If it's easier to find the 3'UTR based on the last exon than to find annotation with UTRs, it's fine.

summary_workflows/identification/specification/I_benchmark_specification.md
daneckaw: Before this version of specifications I realised that TN doesn't really work and removed it from all "obligatory" metrics, but I forgot to update the optional plots. This should be Precision-Recall curve instead of ROC.

summary_workflows/differential_usage/specification/D_benchmark_specification.md
daneckaw: TN can be calculated without problems for differential usage and  FPR metric uses TN. I'll add TN to the list.

summary_workflows/differential_usage/specification/D_benchmark_specification.md
daneckaw: I removed the consolidation file example.

summary_workflows/identification/specification/I_benchmark_specification.md
daneckaw: I removed the consolidation file example.

summary_workflows/quantification/specification/Q_benchmark_specification.md
daneckaw: I removed the consolidation file example.

summary_workflows/quantification/specification/Q_benchmark_specification.md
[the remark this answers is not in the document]
daneckaw: @mzavolan I don't think it has to be MSE, but I think it's useful to have a metric that includes relative usage and not only TPM, as TPM would be impacted if gene expression is over- or underestimated in RNAseq compared to 3'end seq. Do you think it would be better to use MSE for relative usage or to remove relative usage metrics altogether?

@dominikburri 
> calculated as the fraction of most proximal PAS of all PAS

I'm not sure I understand what you mean here, so I'll explain what I had in mind: 1. find all PAS for a given gene, 2. sum TPM values of all PAS, 3. calculate fraction for each PAS by dividing TPM_PAS by TPM_sum. By "all PAS for a given gene" I mean all ground truth PAS for a given gene or all predicted PAS that are matched to ground truth for a given gene.

> Given the above definition, the MSE would measure whether the participant can correctly identify relative PAS usage (or alternative polyadenylation). This is normalised and does not depend on absolute quantification levels, the correlation metric deals with this.

Yes.

> And with gene you mean exonic locations only? Or would you include intronic regions as well for consideration?

If ground truth includes intronic PAS, then predicted PAS in introns will also be counted.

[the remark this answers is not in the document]
dominikburri: Hi @daneckaw, thank you for the explanations! It is clear for me now.

summary_workflows/identification/specification/I_benchmark_specification.md
mzavolan: I agree to leave out the TN and related measures, it's not that informative.

summary_workflows/differential_usage/specification/D_benchmark_specification.md
mzavolan: Yes, for differential polyadenylation, TN is more informative than for identification of PAS.

summary_workflows/identification/specification/I_benchmark_specification.md
```diff
@@ -151,36 +195,44 @@ The OpenEBench assessment file contains the following attributes:
 - **\_id** - follows the format: community:challenge\_metric\_tool
```
ninsch3000, line 195: `[COMMUNITY_ID]:[CHALLENGE_ID]_[PARTICIPANT_ID]_[METRIC_ID]_[WINDOW_SIZE]nt`

summary_workflows/identification/specification/I_benchmark_specification.md
```diff
@@ -151,36 +195,44 @@ The OpenEBench assessment file contains the following attributes:
...
 - **challenge_id** - describing the combination of input dataset with indication whether it is based on real or simulated data and any parameters used for metric calculation; e.g. datasetA\_simulated\_10nt
 - **participant_id** - benchmarked tool
```
ninsch3000, line 197: this is only the challenge name, same as the part [CHALLENGE_ID] in _id

ninsch3000 (changes requested):
Hey @daneckaw , I have again some small comments. I'm sorry if things have changed, it's just everything a work in progress. But as the PR #227 is ready to merge I think it's best to adapt the specs to that status.   
I think more or less everything I say at one aggregation/assessment/gtf and specs file also applies to the other two benchmarking events (except that diff usage is without window sizes).    

One more thing: I think it would be good to add an example of barplot in the Q and I aggregation files. it looks like that:   
```
{
        "_id": "APAeval:2022-04-27_embryonicCortex.PAPERCLIP.mm10_Aggregation_MSE_50nt",
        "challenge_ids": [
            "embryonicCortex.PAPERCLIP.mm10"
        ],
        "datalink": {
            "inline_data": {
                "challenge_participants": [
                    {
                        "metric_value": 0.5,
                        "participant_id": "tool2"
                    }
                ],
                "visualization": {
                    "metric": "MSE_50nt",
                    "type": "bar-plot"
                }
            }
        },
        "type": "aggregation"
    }
```

Thanks a lot again for all the work <3

summary_workflows/identification/specification/I_benchmark_specification.md
```diff
@@ -63,64 +72,72 @@ Fields:
...
+where _d - distance threshold_
 
 Input datasets:
 
 - RNA-Seq data compared with 3'end sequencing data
 - Simulated RNA-Seq data compared with dataset used for simulation
 
+Ranking: The best performing tool is the one with the highest Sensitivity combined with highest Precision (top right part of the plot) and the worst performing tool is the one with the lowest Sensitivity combined with lowest Precision (bottom left part of the plot).
+The plot should be divided into diagonal quartiles based on the distance from optimal performance.
+Alternatively, if the plot is divided into square quartiles, the following ranking order should be applied: top-right, bottom-right, top-left, bottom-left.
```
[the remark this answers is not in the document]
daneckaw: Thanks, I'll change that. I vaguely remember reading that it can be adjusted, but now I cannot find anything like that in the documentation.

summary_workflows/differential_usage/specification/example_files/assessment_out.json
```diff
@@ -0,0 +1,24 @@
+[
+    {
+        "_id": "APAeval:datasetA_TPR_toolA",
```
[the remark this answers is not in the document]
daneckaw: Thanks! I have one question: is the benchmarking event (quantification, identification or differential usage) indicated anywhere in those IDs?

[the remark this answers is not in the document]
ninsch3000: Ohhh, good point... it is not.    
Let me think about how to introduce it, because "benchmarking event" is not a parameter passed to the workflow. I see how this might cause problems though, if in two events the same ground truth and the same metric names are used. On the other hand @AsierGonzalez said that the OEB DB will anyway give a new unique id to each object, and save this verbose id only as additional info. And when querying the DB we use the hardcoded event id (e.g. "OEBE0070000001" for identification) that is part of each summary workflow.
I'll anyway open a new issue to keep that point in mind. And maybe @txellferret also has some thoughts on this? Thanks @daneckaw !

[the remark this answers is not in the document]
daneckaw: At the moment there are different metric names in different benchmarking events (e.g. `Sensitivity_10nt` etc in identification and `Sensitivity` in differential usage) so there shouldn't be any technical issues, but I see how this could be confusing at some point later on.

summary_workflows/identification/specification/example_files/input3.gtf
```diff
@@ -0,0 +1,30 @@
+chr19_GL000209v2_alt	ncbiRefSeq.2021-12-08	transcript	162306	167842	.	-	.	gene_id "LILRP2"; transcript_id "NR_003061.2";  gene_name "LILRP2";
```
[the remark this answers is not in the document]
SamBryce-Smith: Agreed should use Gencode format here as that's what we're using for running execution workflows

summary_workflows/quantification/specification/Q_benchmark_specification.md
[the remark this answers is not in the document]
daneckaw: Thanks for clarification  @mzavolan. I've changed the metric to Correlation instead of MSE.

SamBryce-Smith (commented):
Thanks @daneckaw, sorry for the delayed review! Have only had chance to take a look at identification side in detail. 

General question on the range of distance thresholds - is there any particular reason why we've specified 10, 50 & 100nt as the 'default' thresholds? I see in identification specs there is an optional plot to try 0-200nt in 10nt steps (I like the consistent step-sizes a little more, although there's a risk of information overload especially once get to more lenient windows). I think the particular values are fine, was curious more than anything.

Can I also request a nice but not essential update to the formatting for the 'plots' section in  `summary_workflows/identification/specification/I_benchmark_specification.md`? I find it quite difficult to read at the moment. Perhaps using subheadings in place of the numbering would work better e.g.

\# Plots
\### 1. 2D scatter plot visualizing Sensitivity and Precision of poly(A) site identification.

## Plots
### 1. 2D scatter plot visualizing Sensitivity and Precision of poly(A) site identification.

**Metric**:
**Ranking**:

### 2. bar plot visualizing AUC of Precision-Recall curve with single AUC value for each tool.

summary_workflows/differential_usage/specification/D_benchmark_specification.md
```diff
@@ -0,0 +1,133 @@
...
+
+**X axis** - FPR  
+**Y axis** - TPR
+
+Input datasets:
+
+- RNA-Seq data compared with 3'end sequencing data
+- Simulated RNA-Seq data compared with dataset used for simulation
+
+Ranking: The best performing tool is the one with the highest TPR combined with lowest FPR (top left part of the plot) and the worst performing tool is the one with the lowest TPR combined with highest FPR (bottom left part of the plot. The plot should be divided into diagonal quartiles based on the distance from optimal performance. Alternatively, if the plot is divided into square quartiles, the following ranking order should be applied: top-left, top-right, bottom-left, bottom-right.
```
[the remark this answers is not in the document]
txellferret: If you want to change the best corner, you just need to include it in the aggregation jsons, in the  "visualization": {} object, you can specified by: "optimization": "bottom-right" for example. By default, it is  "optimization": "top-right"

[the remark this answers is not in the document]
ninsch3000: In `assessment_chart.py` it looks like there are only the two options "top-right" and "bottom-right". Is that true for the website viz as well?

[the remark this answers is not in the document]
txellferret: Yes, it is. But we can implement also top-left if necessary :)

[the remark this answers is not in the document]
daneckaw: @txellferret would it be easy/quick to implement it?

Alternatively, I can:
- change the plots to Precision-Recall, which would have top-right as the best corner, and this would give us similar information
- swap axes, and then bottom-right would be the best corner - not the standard way of plotting this, but that's the only issue

summary_workflows/identification/specification/I_benchmark_specification.md
```diff
@@ -16,14 +16,22 @@ Comparison of PAS predicted from simulated RNA-Seq data with dataset used for si
 1. Poly(A) sites identified based on simulated RNAseq data using the benchmarked tool
 2. Ground truth: Poly(A) sites of transcripts used to simulate RNAseq reads
 
+Calculating proportion of identified PAS assigned to different genomic features:
+
+1. Poly(A) sites identified based on RNAseq data using the benchmarked tool
+2. GTF file with genome annotation
+
```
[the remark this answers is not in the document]
daneckaw: Thanks @SamBryce-Smith, that's really useful. I agree it's difficult to use this metric for benchmarking, especially if, as you said, the tools focus on 3'UTRs for identification. 

Your interpretation was correct, the metric wasn't supposed to be calculated for ground truth. I can see how that would be useful just for visualisation, but I think it wouldn't make much sense for OEB benchmarking.

I can entirely remove this metric from OEB plots and move it to optional plots/visualisations for the paper.

summary_workflows/identification/specification/I_benchmark_specification.md
```diff
@@ -63,64 +72,72 @@ Fields:
...
+Input data:
 
 - RNA-Seq data compared with 3'end sequencing data
 - Simulated RNA-Seq data compared with dataset used for simulation
 
-Ranking: The best performing tool is the one with the highest TPR combined with lowest FPR (top left part of the plot) and the worst performing tool is the one with the lowest TPR combined with highest FPR (bottom left part of the plot. The plot should be divided into diagonal quartiles based on the distance from optimal performance. Alternatively, if the plot is divided into square quartiles, the following ranking order should be applied: top-left, top-right, bottom-left, bottom-right.
+Ranking: The best performing tool is the one with the highest AUC value.
 
-3. **bar plot** visualizing **AUC** of ROC plot with single AUC value for each tool.
+3. **bar plot** visualizing **proportion of PAS assigned to 3'-UTRs**.
```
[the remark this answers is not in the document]
daneckaw: The only point of having those separately was that while we could rank tools based on proportion assigned to 3'UTRs in OEB (though I agree it might not be the best metric, or maybe it should be calculated in relation to ground truth), I don't think we could rank it for all other features. And so proportion in 3'UTRs would be part of OEB plots, and proportion in different features (including 3'UTR) would be part of additional plots made only for the paper and not for ranking the tools.

summary_workflows/identification/specification/I_benchmark_specification.md
```diff
@@ -63,64 +72,72 @@ Fields:
...
 
-**X axis** - benchmarked tool  
-**Y axis** - AUC
+**metric** - proportion of PAS assigned to 3'-UTRs
 
 Input data:
 
 - RNA-Seq data compared with 3'end sequencing data
 
+Ranking: The best performing tool is the one with the highest proportion of PAS assigned to 3'-UTRs
```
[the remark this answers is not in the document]
daneckaw: Thanks, that was something I wasn't sure about myself. I was also thinking of something along the lines of "in the 3'-UTR or a reasonable distance downstream from 3'UTR", but that's getting more vague. It will be easier to just add the plots without ranking and leave the interpretation to the reader.

summary_workflows/identification/specification/I_benchmark_specification.md
```diff
@@ -130,6 +147,32 @@ Ranking: For each tool, the area under the curve is calculated. The best perform
 
 Note: 2D line plot is not supported in OpenEBench yet. If it's not implemented, the data should be visualised outside of OpenEBench.
 
+2. Grouped **bar plot** visualizing **proportions of PAS assigned to different genomic features** in one plot.
+
+**metric** - proportions of PAS assigned to different genomic features. Multiple bars corresponding to multiple featues are plotted using different colors.
```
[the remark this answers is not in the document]
daneckaw: This is more or less the plot format I had in mind:

example_plot
- intron
- everything else

Based on your comments on 3'UTR metric, I'd also add ground truth to the plots along the benchmarked tools.

summary_workflows/identification/specification/example_files/input3.gtf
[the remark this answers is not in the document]
daneckaw: Thanks, I'll change that.

summary_workflows/identification/specification/I_benchmark_specification.md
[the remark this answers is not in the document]
SamBryce-Smith: I see that sounds sensible, especially as the majority of tools only consider polyA sites within 3'UTRs

summary_workflows/identification/specification/I_benchmark_specification.md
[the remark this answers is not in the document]
SamBryce-Smith: Makes sense with the example plot thank you! I like the categories you suggest, would definitely have the 'exon' category exclude the 3'-terminal exon. I think it would be nice to have an additional category that is 'downstream of 3'-terminal exon' within an arbitrary distance (e.g. 1kb, 5kb etc.). I don't think that is essential though, 'intergenic' for the rest would cover just fine for now.

summary_workflows/identification/specification/I_benchmark_specification.md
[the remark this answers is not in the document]
SamBryce-Smith: Looks like this has been removed with 6d41516f4095c21744cd9e03bc001e80ba0b8466 so closing

summary_workflows/identification/specification/I_benchmark_specification.md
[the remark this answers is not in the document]
dominikburri: Do we still need to do something here?

dominikburri (dismissed):
Thank you very much @daneckaw for all the work!

I went over the specs again and I think it is fine, all the plots for OEB make sense I think. 
If the remaining open questions and comments can be marked as "resolved" that would be perfect.
Aside from this, happy to merge it.

daneckaw:
Thanks for reviewing @dominikburri. I've added some final edits and it should be ready for merging now.

Changes:
- input GTF files changed to Gencode format
- formatting of plot descriptions to improve readability
- description of optional identification plots and metrics (proportion assigned to genomic features)

summary_workflows/identification/specification/I_benchmark_specification.md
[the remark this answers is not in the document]
daneckaw: Sorry @SamBryce-Smith, I missed your question about window size/distance threshold. So ideally I would prefer to have a wider range (eg from 0 to 200nt), but given limitations of how many separate 2D scatter plots we can reasonably show, I thought that a few values closer to the ground truth might be more informative. I still think it might be useful to plot up to 200nt in a single 2D line plot, so I left it like that in optional plots.

summary_workflows/identification/specification/I_benchmark_specification.md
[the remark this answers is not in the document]
daneckaw: I've changed the features to:
- 3'-terminal exons
- exon (excluding 3'-terminal exon?)
- introns
- intergenic region within 1kb downstream of 3'-terminal exon
- intergenic region more than 1kb downstream of 3'-terminal exon

This is an optional plot, so if we think something else might be more useful later on, we can still change it.

summary_workflows/identification/specification/I_benchmark_specification.md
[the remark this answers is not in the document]
SamBryce-Smith: Agreed, think it's a good idea to keep every possible step to the optional plots, especially if we can't visualise all the lines (for every window size) on a single plot in OpenEBench!

summary_workflows/quantification/specification/Q_benchmark_specification.md
```diff
@@ -65,44 +92,73 @@ Fields:
...
-Ranking: The best performing tool is the one with the highest correlation value.
+### 2. Bar plot visualizing Correlation of relative PAS usage. 
+
+**Plot type**: Bar plot
 
-Optional plots:
+**Metric**: Correlation of relative PAS usage
 
-2.  **2D line plot** visualising **correlation** as a function of distance threshold.
+**Ranking**: The best performing tool is the one with the highest correlation value value.
```
SamBryce-Smith, line 132: Don't need the second 'value'. Not a big deal though

summary_workflows/identification/specification/I_benchmark_specification.md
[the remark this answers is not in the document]
SamBryce-Smith: This is ok on my end now!

PR #235, opened 2022-04-26
Ag fix vre vis

AsierGonzalez:
I had to do some small changes so that the visualisations produced by the summary workflow are visible in the VRE. The main change is that in the `results/{challenge_id}` directory there will be as many `{challenge_id}.json` files as aggregations. To be accurate, the first one will be `{challenge_id}.json` and the rest will have a number after the `{challenge_id}.json`. The VRE will visualise the results in the first JSON only, this will be fixed in the future.

## Checklist

- [x] I have tested my code

### DIFF
--- a/summary_workflows/quantification/quantification_dockers/q_consolidation/aggregation.py
+++ b/summary_workflows/quantification/quantification_dockers/q_consolidation/aggregation.py
@@ -11,7 +11,7 @@ from OEB_aggr_query import OEB_aggr_query
 # set to production link when ready
 DEFAULT_OEB_API = "https://dev-openebench.bsc.es/api/scientific/graphql"
 # Make sure to adapt accordingly in other event workflows; Here is APAeval:Quantification
-DEFAULT_bench_event_id = "OEBE0070000002" #  identification: "OEBE0070000001", differential usage: "OEBE0070000003
+DEFAULT_bench_event_id = "OEBE0070000000" # New benchmarking events (still empty): identification: "OEBE0070000001", quantification: "OEBE0070000002", differential usage: "OEBE0070000003"
 
 class Visualisations(Enum):
     """Visualisations supported for plotting.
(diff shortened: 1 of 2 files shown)


summary_workflows/quantification/quantification_dockers/q_consolidation/aggregation.py
```diff
@@ -144,10 +144,17 @@ def main():
 
         logging.debug(f"aggregation after update: {new_aggregation}")
 
-        # 2.e) Write aggregation file per challenge to local results dir
-        aggregation_file = os.path.join(challenge_dir, challenge_id + ".json")
-        with open(aggregation_file, mode='w', encoding="utf-8") as f:
-            json.dump(new_aggregation, f, sort_keys=True, indent=4, separators=(',', ': '))
+        # 2.e) Write each aggregation object to a separe file per challenge to local results dir
+        for count, aggr_object in enumerate(new_aggregation):
```
ninsch3000, line 148: @AsierGonzalez I'm afraid with this approach the `consolidated_result.json` will also only contain the first aggregation for each challenge. However, if I understood correctly, the web visualization of OEB *can* handle multiple aggregations per challenge and thus create multiple plots per challenge. Actually the whole exercise of having the branch `swf_multiple_plots` was to be able to submit a `consolidated_result.json` with *multiple* aggregation objects per challenge to the OEB DB.    
As I see it we have several options:   
- discard the VRE visualization and just use the web viz, as it was without the present PR
or
- adapt the script `merge_data_model_files.py` to loop over all those numbered aggregation files that you create, so that they are united again in the `consolidation_results.json`
or
- write the aggregation file with multiple objects *in addition* to the numbered ones, and only load that "multiple objects file" in the `merge_data_model_files.py`

ninsch3000 (changes requested):

summary_workflows/quantification/quantification_dockers/q_consolidation/aggregation.py
AsierGonzalez: I think I fixed it.

dominikburri (commented):
Hi @AsierGonzalez, thank you for the fixes!

I tested the workflow with local docker images (rebuilt as described in the readme) and `offline=1`. This generated your mentioned files in `results`: `Qx.json` and `Qx-1.json`. 

Using `offline=0` leaves the `results/` empty. I think this is because no file is available on the DB and the templates are not used from there. 
Would we not want to use the template `aggregation_template_Q.json` if no previous aggregation object from the OEB DB? 
Do you get the same output? 

If I remove `Qx.json` (previous aggregation) and set  `offline=1`, the template is used and I get the aforementioned files in `results/`. 

I think once this question is solved, it can be merged.

dominikburri:
> Hi @AsierGonzalez, thank you for the fixes!
> 
> I tested the workflow with local docker images (rebuilt as described in the readme) and `offline=1`. This generated your mentioned files in `results`: `Qx.json` and `Qx-1.json`.
> 
> Using `offline=0` leaves the `results/` empty. I think this is because no file is available on the DB and the templates are not used from there. Would we not want to use the template `aggregation_template_Q.json` if no previous aggregation object from the OEB DB? Do you get the same output?
> 
> If I remove `Qx.json` (previous aggregation) and set `offline=1`, the template is used and I get the aforementioned files in `results/`.
> 
> I think once this question is solved, it can be merged.

After discussion with @AsierGonzalez and @ninsch3000 , this behaviour is still missing, i.e. using template when empty DB. It will be fixed in the merging branch #227. Thus, these fixes can be merged as is.

PR #238, opened 2022-04-29
feat: quantification metric: relative PAS usage

### LINKED ISSUE #231: Summary workflows: add script for MSE calculation
A function for calculation of MSE is still missing in the Quantification summary workflow, we currently only have a placeholder.

dominikburri:
Fixes #231 

Compute the missing MSE metric. 
Right now, the correlation of relative PAS usage (per gene) is computed. The correlation can be replaced by MSE though, see discussion in #222.
Also, the correct computation depends on #228, which should be merged into this one before merging this PR. 
Therefore, still a draft.

The nextflow worklow has an additional parameter: `genome_file`, which is a genome annotation file in gtf format. This is used for getting the relative PAS usage per gene. Then, the correlation on these relative values is computed. 

> This increases the runtime considerably. Might be able to increase the performance.

This is pretty much fixed in commit 69280e8, where the metric computation is only performed for the first window size.

## Checklist

- [x] I have tested my code


dominikburri:
Missing:

- [ ] If needed, change correlation to MSE metric.
- [ ] rename metric name from *MSE* to what is in the specification.
- [ ] Test the correct computation of the relative PAS usage with simple cases.

dominikburri:
Hi @txellferret, I changed the genome parameter to use a directory and renamed the test data accordingly. I could only quickly test it and if you could try again on the VRE that would be great.
The main change is that the `challenge` now needs the organism specifier, I documented this in the readme. 

Hi @SamBryce-Smith and @faricazjj, unfortunately I don't have time in the next days to implement the missing points in [this comment](https://github.com/iRNA-COSI/APAeval/pull/238#issuecomment-1117117841). Would one of you be able to take care of it? That would be great.

txellferret (approved):
Tested workflow and the execution is working successfully.

SamBryce-Smith:
I addressed one of the points in @dominikburri's [previous comment](https://github.com/iRNA-COSI/APAeval/pull/238#issuecomment-1117117841) RE updating variable names from MSE to correlation (of relative usage). I couldn't see whether we have specified variable names so went with names I thought were sensible. I updated the metric name reported in the JSON as per https://github.com/iRNA-COSI/APAeval/commit/6afd33625c10064132f5696e9b176189670381d4. Want to note that this seems to be the only metric ID that is capitalised (for 1st letter), maybe that is/isn't an issue. Just wanted to follow the specifications exactly.

Got a successful local run with test data following instructions in the README
image

summary_workflows/quantification/quantification_dockers/q_consolidation/aggregation_template_Q.json
```diff
@@ -9,7 +9,7 @@
                 "challenge_participants": [],
                 "visualization": {
                     "type": "2D-plot",
-                    "x_axis": "percent_matched",
+                    "x_axis": "expression_unmatched",
```
daneckaw, line 12: I made a mistake when changing the scripts to deal with matched/non_matched sites and used `expression_unmatched` as a name for the new metric. Could you change it in this PR to `Expression_non-matched-PAS`? I think that "non-matched" makes more sense than "unmatched" in this case.

SamBryce-Smith (dismissed):
Thanks for recent changes and the bump @dominikburri ! As per 88bf4f586ed16ea2a11fbf80685398157e603ccf I think the relative usage is now calculated appropriately.

Tried another test run locally and was successful. As someone who isn't very familiar with the summary workflows this was very straightforward for me to run (kudos to you all for test data and documentation!).

Not a blocking comment for me but I think it would be good down the line to try and get the relative usage correlation working for all window sizes. As it stands calculating for the most permissive window does the job nicely though


Looks good to go on my end :) I can't comment too specifically about the OpenEBench side, but from a browse of the outputs everything seems to be in order.

quant-summary-workflow-test2022-05-17_16-06-03

dominikburri:
> Not a blocking comment for me but I think it would be good down the line to try and get the relative usage correlation working for all window sizes. As it stands calculating for the most permissive window does the job nicely though

Thank you @SamBryce-Smith for reviewing! 
Indeed, we first had it that it would calculate relative PAS usage for every provided window size. But since we don't need these values for plots on OEB and the calculations are more time intensive, we decided to calculate the metric only for the first window size.

I tested now also the workflow with a sample with preceding "chr". For this, I manually changed the sample bed, ground truth bed and gtf to use "chr" in the sequence names. Everything worked and the results (metric values in *assessment* types) were identical.

summary_workflows/quantification/README.md
```diff
@@ -28,6 +28,12 @@ This README describes the APAeval **quantification** summary workflow. For a mor
 - `input_file`: output file from execution workflow in bed6 format
 - `gold standard`: bed6 file derived from 3'end sequencing on the same sample(s) as the RNA-seq data used in the challenge
 >NOTE: the gold standard file MUST be named in the format `[challenge].bed`, where `[challenge]` is specified in `challenges_ids` in `nextflow.config`. The extension `.bed` is hardcoded within [`compute_metrics.py`][metrics-py].
+- `windows` parameter is used to compute metrics for a list of window sizes.
+    - For running on OEB: the parameter is read from `nextflow.config`.
+- `genome_dir`: Directory to genome annotation in gtf format with 9 fields as specified [here](https://www.ensembl.org/info/website/upload/gff.html). The gtf is used for the relative PAS usage metric computation.
```
ninsch3000, line 33: Given the fact that we want to use Gencode annotations in APAeval, and are already having lot's of confusion with Gencode vs Ensembl I think it would make sense here to give a link to Gencode; https://www.gencodegenes.org/pages/data_format.html  
:smile:

ninsch3000 (dismissed):

SamBryce-Smith (approved):
LGTM! Good spot on changing the test data to Gencode chromosome format I should have double-checked that more closely...

daneckaw (approved):
Looks good!

PR #239, opened 2022-05-03
feat: identification summary workflow

ninsch3000:
The quantification summary workflow and associated files have been copied into the identification directory. Filenames have been adapted.
- validation should be the same as in quantification; currently only checks for presence of bed columns (not checking content of score col) are performed, as well as a comparison of the chromosome naming format between input and gtf file (have to be both with or without leading "chr")
- compute_metrics has to be adapted: 
  - function definitions should be written in `identification_metrics.py`. There, the functions from the quantification metric calculation are also still present, those should be either adapted if needed, or otherwise deleted from the script. 
  - function calls have to be adapted in `compute_metrics.py`. Currently the calls are commented out and instead the variables holding the metrics are set to arbitrary values, in order to test the workflow (delete those)
- consolidation should be the same as in quantification.
-  test files might have to be adapted to identification specific cases

The workflow runs through on my Ubuntu 20.04 machine with the specified dummy metric values. 

Fixes nothing

## Checklist

- [ ] I have tested my code


ninsch3000:
Please see PR description for instructions

daneckaw:
Tests for metrics calculations:
[identification_metrics_test_report.pdf](https://github.com/iRNA-COSI/APAeval/files/9050228/identification_metrics_test_report.pdf)

PR #243, opened 2022-05-06
fix: incompatible col types in matchPAS

ninsch3000:
In `matchPAS.py` the function `bedtools_window` returned a dataframe, where the predicted PAS chromosome was type `str`, and the ground truth chromosome `int`. Consequently, subsetting in subsequent functions was erroneous.

Fix: return both chromosome columns as type `str`.


faricazjj (approved):
Thanks for the fix @ninsch3000 ! :D

ninsch3000:
Ok, it wouldn't matter if we finally used the gencode chr naming scheme consistently, but better save than sorry :sweat_smile:  

That's a reminder to update all test files to gencode convention.

SamBryce-Smith (approved):
Looks good to me thanks @ninsch3000! As you say I think it's better to be explicit here even if it won't be a problem with all gencode-formatted files.

PR #256, opened 2022-05-24
Update Quantification event ID

ninsch3000:
Update Quantification event ID from OEBE007000000 to ...2, as Asier had set new IDs. Please check with the OEB DB whether this ID is now correct.


txellferret:
Also the api enpoint should be: DEFAULT_OEB_API = "https://openebench.bsc.es/api/scientific/graphql"
to point and fetch data from the production db

txellferret (approved):
Execution is working successfully for test data. 
image

dominikburri (approved):
Thanks @ninsch3000 and @txellferret!
Didn't test, since the changes shouldn't introduce problems on the local side but only with OEB integration. Looks correct from my side.

PR #259, opened 2022-05-27
feat: Validation with whole gtf

### LINKED ISSUE #258: Validation fails with mitochondrial genes
The current implementation of validating the input BED checks whether the genome annotation is with "chr" or without, starting here: 
https://github.com/iRNA-COSI/APAeval/blob/474f80976e950f78e677f4c1852ac3e4dab4fec4/summary_workflows/quantification/quantification_dockers/q_validation/validation.py#L66

The set of accepted chromosome names is manually constructed, starting here:
https://github.com/iRNA-COSI/APAeval/blob/474f80976e950f78e677f4c1852ac3e4dab4fec4/summary_workflows/quantification/quantification_dockers/q_validation/validation.py#L114

This fails when the input BED contains "chrM" or "M" or some derivation for the mitochondrial genes. 

The quick fix would be to not allow mitochondrial genes in the input. But I don't think this is the intended goal. 

This affects all benchmarking events, not only quantification which I mention here. That is: #239.

## Fix

- [x] Obtain set of chromosome names directly from genome annotation. Requires reading all lines. 
- [x] Compare input BED chromosomes against the annotation set. 
- [ ] Update docker images on VRE and dockerhub.

dominikburri:
Improve validation in quantification summary workflow by comparing the chromosome names from the input against the list of chromosome names from the appropriate genome annotation. 
The genome annotation is now completely read, not only the first `n` lines. The code runs not much longer, I tested it on `gencode.hg38.v38.annotation.gtf`.

Fixes #258 

## Checklist

- [x] I have tested my code

### DIFF
--- a/summary_workflows/quantification/README.md
+++ b/summary_workflows/quantification/README.md
@@ -19,7 +19,7 @@ This README describes the APAeval **quantification** summary workflow. For a mor
    - input file has to be tab separated file with 6 columns
    - start and end coordinates (col 2,3) have to be int64
    - strand (col 6) has to be one of [+,-]
-   - chromosome (col 1) has to be one of [1..22,X,Y]   
+   - chromosome (col 1) has to match the ones from the genome annotation (see below `genome_dir`)
   
 - The `validated-participant-data.json` file is not used in the subsequent steps, but the workflow exits if the input file doesn't comply to the specifications of the current benchmarking event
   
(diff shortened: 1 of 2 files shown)


yuukiiwa (approved):
Thanks for improving the validation, @dominikburri! I think this is a lot better!! :)

faricazjj (approved):
LGTM thanks @dominikburri ! :D

PR #328, opened 2022-06-21
include q metrics with different return df types

### LINKED ISSUE #325: Quantification summary workflow: Add without unmatched site option in metric calculation
Based on the reverted commit: https://github.com/iRNA-COSI/APAeval/pull/262/commits/deaf84abbcf3a56e4283b0524fec6126b055a4f4

dominikburri:
Fixes #325 

## Checklist

- [x] I have tested my code

### DIFF
--- a/summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py
+++ b/summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py
@@ -81,7 +81,8 @@ def compute_metrics(participant_input, gold_standards_dir, challenge_ids, partic
             if window == windows[0]:
                 match_with_gt_and_pd = matchPAS.match_with_gt(participant_input,gold_standard,window, "with_unmatched_GT_and_PD")
                 merged_bed_df_wpd, _ = match_with_gt_and_pd[0], match_with_gt_and_pd[1]
-                correlation_rel_use = matchPAS.relative_pas_usage(merged_bed_df_wpd, genome)
+                normalised_df = matchPAS.relative_pas_usage(merged_bed_df_wpd, genome)
+                correlation_rel_use = matchPAS.corr_with_gt(normalised_df)
                 # Key: exact name of metric as it appears in specification
                 metric_name = f"Correlation_coefficient_relative_{window}nt"
                 # Value: List of [variable_holding_metric, std_err]
--- a/summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py
+++ b/summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py
@@ -55,19 +55,18 @@ def compute_metrics(participant_input, gold_standards_dir, challenge_ids, partic
 
         for window in windows:
             for return_df_type in all_return_df_types:
-
-                # METRIC: Matched sites
-                ########################
-                # metric on the number of matched sites
+                # obtain dataframe for metric computation
                 match_with_gt_run = matchPAS.match_with_gt(f_PD=participant_input, f_GT=gold_standard, 
                     window=window, return_df_type=return_df_type)
                 merged_bed_df, expression_unmatched = match_with_gt_run[0], match_with_gt_run[1]
-
+                
+                # METRIC: Matched sites
+                ########################
+                # metric on the expression of non matched polyA sites.
                 # Key: exact name of metric as it appears in specification
                 metric_name = f"Expression_non-matched-PAS_{window}nt"
                 # Value: List of [variable_holding_metric, std_err]
                 metrics[metric_name] = [expression_unmatched, 0]
-
 
                 # METRIC: correlation coefficient
                 #################################
@@ -88,10 +87,7 @@ def compute_metrics(participant_input, gold_standards_dir, challenge_ids, partic
                 ####################
                 # Only calculate metric for first window size.
                 if window == windows[0]:
-                    match_with_gt_and_pd = matchPAS.match_with_gt(f_PD=participant_input, f_GT=gold_standard,
-                        window=window, return_df_type=return_df_type)
-                    merged_bed_df_wpd, _ = match_with_gt_and_pd[0], match_with_gt_and_pd[1]
-                    normalised_df = matchPAS.relative_pas_usage(merged_bed_df_wpd, genome)
+                    normalised_df = matchPAS.relative_pas_usage(merged_bed_df, genome)
                     correlation_Pearson_rel_use = matchPAS.corr_Pearson_with_gt(normalised_df)
                     # Key: exact name of metric as it appears in specification
                     metric_name = f"Correlation_coefficient_Pearson_relative_{return_df_type}_{window}nt"
--- a/summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py
+++ b/summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py
@@ -36,6 +36,8 @@ def compute_metrics(participant_input, gold_standards_dir, challenge_ids, partic
 
     # define array that will hold the full set of assessment datasets
     all_assessments = []
+    # define list with keywords for return type of dataframe from match_with_gt()
+    all_return_df_types = ["with_unmatched_GT", "with_unmatched_GT_and_PD", "without_unmatched"]
 
     for challenge in challenge_ids:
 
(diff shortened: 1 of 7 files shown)


ninsch3000 (approved):
Not tested but LGTM

faricazjj (dismissed):
LGTM :D

summary_workflows/quantification/specification/Q_benchmark_specification.md
```diff
@@ -190,15 +190,16 @@ The following tables list the metric names, value types and units, and a descrip
...
+| `Correlation_coefficient_Pearson_with_unmatched_GT_100nt` | `float` | N/A | Correlation between poly(A) site quantification by benchmarked tool and ground truth dataset calculated for 100 nt distance threshold (window) |
 | `Expression_non-matched-PAS_10nt` | `float` | N/A | Total expression of non-matched sites quantified by benchmarked tool for 10 nt distance threshold (window) |
 | `Expression_non-matched-PAS_50nt` | `float` | N/A | Total expression of non-matched sites quantified by benchmarked tool for 50 nt distance threshold (window) |
 | `Expression_non-matched-PAS_100nt` | `float` | N/A | Total expression of non-matched sites quantified by benchmarked tool for 100 nt distance threshold (window) |
-| `Correlation_coefficient_Pearson_relative_100nt` | `float` | N/A | Correlation between relative PAS usage calculated from RNAseq-based PAS quantification and orthogonal 3'end seq data for 100 nt distance threshold (window) |
+| `Correlation_coefficient_Pearson_relative_with_unmatched_GT_and_PD_100nt` | `float` | N/A | Correlation between relative PAS usage calculated from RNAseq-based PAS quantification and orthogonal 3'end seq data for 100 nt distance threshold (window) |
```
faricazjj, line 200: Are we possibly missing the description of without unmatched in this table?

faricazjj (commented):
Thank you @dominikburri ! :D I just have one question, everything else looks good!

summary_workflows/quantification/specification/Q_benchmark_specification.md
dominikburri: Mh in this table in the specification only metrics are described that appear in the plots. The note below mentions that all three versions are computed and available in the `assessment_dataset.json` and in the `consolidated_result.json`.
My intend was to keep the computations as they were but include the additional ones in the json files. Once we settle on the metric to plot and to use, we can adjust it here in the specification. 
Do you agree?

faricazjj: Ah i see! That makes sense! I thought it was a list of the metrics we have now hehe if that's the case then I was thinking maybe we could add somewhere a consolidated list of the possible metric calculation options and their descriptions, but I guess that should be a separate PR :D

faricazjj (approved):
LGTM thank you @dominikburri !!

summary_workflows/quantification/specification/example_files/assessment_out.json
```diff
@@ -1,46 +1,46 @@
 [
```
ninsch3000, line 1: I'm afraid this example file is not correct anymore if you loop over `windows` AND `return_df_types` in `compute_metrics.py`. However, it would be quite an overhead to specify all metrics combinations. So maybe you can add some "..." and specify at least each return type once. Or instead of writing the exact id/name, write it with [METRIC_NAME] and [RETURN_TYPE] or sth like that.

summary_workflows/quantification/specification/Q_benchmark_specification.md
```diff
@@ -190,15 +190,16 @@ The following tables list the metric names, value types and units, and a descrip
...
 
 > The correlation metrics above are also computed for Spearman correlation, but currently not used in the plots. The metric ids are adjusted accordingly.
+> The correlation metrics above are besides *with_unmatched_GT* and *with_unmatched_GT_and_PD* for relative correlation also computed for *with_unmatched_GT_and_PD*, *without_unmatched* and *with_unmatched_GT*. That is, all three variants are computed, but only the ones above are plotted. The metric ids are adjusted accordingly.
```
ninsch3000, line 202: Could you rephrase that? E.g.   
"The above mentioned correlation metrics are actually calculated for three different matching types - `with_unmatched_GT`, `with_unmatched_GT_and_PD` and `without_unmatched`, and for Pearson's as well as Spearman's r. All these variants will appear in the `assessment_out.json` and `consolidated_results.json`. However, the `aggregation_out.json` will only contain the metrics explicitly specified for visualization in `aggregation_template.json`"

summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py
```diff
@@ -52,53 +54,55 @@ def compute_metrics(participant_input, gold_standards_dir, challenge_ids, partic
...
-            # Only calculate metric for first window size.
-            if window == windows[0]:
-                match_with_gt_and_pd = matchPAS.match_with_gt(participant_input,gold_standard,window, "with_unmatched_GT_and_PD")
-                merged_bed_df_wpd, _ = match_with_gt_and_pd[0], match_with_gt_and_pd[1]
-                normalised_df = matchPAS.relative_pas_usage(merged_bed_df_wpd, genome)
-                correlation_Pearson_rel_use = matchPAS.corr_Pearson_with_gt(normalised_df)
+                # METRIC: Matched sites
+                ########################
+                # metric on the number of matched sites
+                match_with_gt_run = matchPAS.match_with_gt(f_PD=participant_input, f_GT=gold_standard,
```
ninsch3000, line 62: With the refactoring you introduced (returning the df here and the passing it to the Pearson and Spearman functions), this call should actually go above "METRIC: Matched sites". The return of `match_with_gt` is used in the three subsequent metric calculations

summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py
```diff
@@ -52,53 +54,55 @@ def compute_metrics(participant_input, gold_standards_dir, challenge_ids, partic
...
+                metric_name = f"Correlation_coefficient_Spearman_{return_df_type}_{window}nt"
                 # Value: List of [variable_holding_metric, std_err]
-                metrics[metric_name] = [correlation_Spearman_rel_use, 0]
+                metrics[metric_name] = [correlation_spearman, 0]
+
+                # METRIC: correlation coefficient of relative pas usage
+                ####################
+                # Only calculate metric for first window size.
+                if window == windows[0]:
+                    match_with_gt_and_pd = matchPAS.match_with_gt(f_PD=participant_input, f_GT=gold_standard,
```
ninsch3000, line 91: the variable names `match_with_gt_and_pd`, as well as `merged_bed_df_wpd` are misleading after the refactoring you did, because now this df is calculated for all three return types. Actually, couldn't you therefore even use the `merged_bed_df` from above? As far as I understand the code there is no need to re-calculate it.

ninsch3000 (changes requested):
Thanks a lot @dominikburri . I think it would be nice to further clean up this PR, see my individual comments.    
Moreover, should we maybe create a new issue to make shorter metric names? Like `Pearson_r` instead of `Correlation_coefficient_Pearson_...`.   
After you address my comments I'd still want to test the scripts with the small example file to check the calculations. Just let me know when it's ready :pray:

summary_workflows/quantification/specification/example_files/assessment_out.json
dominikburri: I cannot add comments in json files. So I can add all metrics that are also in the example, in total 27.

summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py
dominikburri: good catch! Will change

dominikburri:
> Thanks a lot @dominikburri . I think it would be nice to further clean up this PR, see my individual comments.
> Moreover, should we maybe create a new issue to make shorter metric names? Like `Pearson_r` instead of `Correlation_coefficient_Pearson_...`.
> After you address my comments I'd still want to test the scripts with the small example file to check the calculations. Just let me know when it's ready 🙏

Thanks @ninsch3000, I addressed your comments. 
As for the metric names, I think a new PR together with @faricazjj's suggestion of more detailed description should be in a separate PR.

faricazjj (approved):
LGTM thanks @dominikburri ! :D

PR #331, opened 2022-06-29
add percent unmatched

### LINKED ISSUE #327: Quantification summary workflow: add percent unmatched

dominikburri:
add metric Percent_non-matched-PAS to assessment,
add FP, FN and TP to restructured return value as dict
(instead of float).

Fixes #327 

## Checklist

- [x] I have tested my code

### DIFF
--- a/summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py
+++ b/summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py
@@ -56,14 +56,16 @@ def compute_metrics(participant_input, gold_standards_dir, challenge_ids, partic
             # METRIC: Matched sites
             ########################
             # metric on the number of matched sites
-            match_with_gt_run = matchPAS.match_with_gt(participant_input,gold_standard,window)
-            merged_bed_df, expression_unmatched = match_with_gt_run[0], match_with_gt_run[1]
+            match_with_gt_run = matchPAS.match_with_gt(participant_input,gold_standard,window,return_df_type="with_unmatched_GT_and_PD")
+            merged_bed_df, summary_statistics = match_with_gt_run[0], match_with_gt_run[1]
 
             # Key: exact name of metric as it appears in specification
             metric_name = f"Expression_non-matched-PAS_{window}nt"
             # Value: List of [variable_holding_metric, std_err]
-            metrics[metric_name] = [expression_unmatched, 0]
-
+            metrics[metric_name] = [summary_statistics["nonmatched_expression"], 0]
+            # METRIC: Percent unmatched, based on TPM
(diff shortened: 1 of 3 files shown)


summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py
```diff
@@ -56,14 +56,16 @@ def compute_metrics(participant_input, gold_standards_dir, challenge_ids, partic
...
 
             # Key: exact name of metric as it appears in specification
             metric_name = f"Expression_non-matched-PAS_{window}nt"
             # Value: List of [variable_holding_metric, std_err]
-            metrics[metric_name] = [expression_unmatched, 0]
-
+            metrics[metric_name] = [summary_statistics["nonmatched_expression"], 0]
+            # METRIC: Percent unmatched, based on TPM
+            metric_name = f"Percent_non-matched-PAS_{window}nt"
+            metrics[metric_name] = [summary_statistics['percent_nonmatched'], 0]
```
ninsch3000, line 68: I think we need a better name here as this sounds a lot like looking at number of PAS.I'll add this to #333

ninsch3000 (approved):
LGTM. Thanks @dominikburri

summary_workflows/quantification/quantification_dockers/q_metrics/matchPAS/matchPAS.py
```diff
@@ -173,13 +173,23 @@ def match_with_gt(f_PD, f_GT, window, return_df_type = "with_unmatched_GT"):
...
-        
-    return(out_df, nonmatched_expression)
+        nonmatched_expression = 0.0
+        percent_nonmatched = 0.0
+
+    summary_statistics = {"nonmatched_expression": nonmatched_expression,
+        "percent_nonmatched": percent_nonmatched,
+        "n_nonmatched_PD": out_rev_PD.shape[0], # FP
+        "n_nonmatched_GT": out_rev_GT.shape[0], # FN
+        "n_matched": out.shape[0]} # TP
```
daneckaw, line 190: I think that for `out.shape[0]` to be considered the number of TP, we need to check if the expression of PD is non-zero. If a site is reported based on the reference database used by the tool, but quantification for this site is zero, then this is a FN.

daneckaw: If we want to use any of those 3 (`n_nonmatched_PD`, `n_nonmatched_GT` or `n_matched`) then it should be changed. But maybe not necessarily in this PR.

dominikburri: I removed them for now, still can add it back in another PR with better computation.

ninsch3000 (approved):

PR #334, opened 2022-06-29
Fix: bug matchPAS 

### LINKED ISSUE #305: bug: python object passed as reference
https://github.com/iRNA-COSI/APAeval/blob/c9e40a9672d5b98e7f6fc2c4c3aa6d7fc18c0fcb/summary_workflows/quantification/quantification_dockers/q_metrics/matchPAS/matchPAS.py#L54

The function obtains variable `matched_sites` but it appears the global variable is used and thus the results are not valid (seen as exceptionally high TPM values for certain PAS, because the same first two rows are used for the same object within the lambda statement in line 89). 

The fix could involve calling the function in line 48 with `matched_sites_new` and adding `matched_sites = matched_sites_new.copy()` prior to line 54. 

Maybe this also happens in other functions, like lines 67, 101, 179, but not sure.

ninsch3000:
Fixes #305

I included the new small test files with which I was able to reproduce the bug. Too high tpms were calculated for PD sites that match to multiple GT, and additionally some of those sites were lost. The fix suggested by @dominikburri solved those problems.

## Checklist

- [ ] I have tested my code


summary_workflows/quantification/APAeval_Q_test_data/metrics_ref_datasets/test.gt.mm10_test.bed
```diff
@@ -0,0 +1,14 @@
+chr1	1000000	1000001	same_pd_21	20	+
```
dominikburri, line 1: should it be this file: `summary_workflows/quantification/APAeval_Q_test_data/metrics_ref_datasets/test.mm10_test.gt.bed`? 
If this is a new one and should be used in the test, please change it in `nextflow.config` to use it, see my other comment.

summary_workflows/quantification/APAeval_Q_test_data/test.pd.mm10_test.bed
```diff
@@ -1,10 +1,12 @@
-chr1	2199734	2199735	not_in_pd	156.240540124	-
```
dominikburri, line 1: renaming the file is fine, but you need to adjust the config such that the test runs as expected. `https://github.com/iRNA-COSI/APAeval/blob/6a00c0428c4c8e970e17e06c3739328a0ea0a935/summary_workflows/quantification/nextflow.config#L40`

daneckaw:
I'm not sure why but I can't replicate the bug outside of the summary workflow, but the fix works as expected within the workflow with the test files provided. I agree with the comments by @dominikburri on the names of the test files and changes in `nextflow.config` but otherwise it looks fine.

dominikburri (approved):
Thank you, ran the local test without issues.
LGTM.

PR #373, opened 2022-07-07
Add relative usage quantification files to execution workflow output file specifications

### LINKED ISSUE #372: Update execution workflow output file specifications to report fractional relative usage values
As in title, we want to evaluate the relative usage values reported by tools and we need to capture them into standardised output formats for consumption by summary workflows.

Specifications (`execution_workflows/execution_output_specification.md`) need to be updated before we can modify the execution workflows.

SamBryce-Smith:
Fixes #372

## Checklist

- [ ] I have tested my code

Suggest 2 new output files depending on the metrics calculated by the tool:
1. Output file to report per-PAS fractional relative usage (if applicable). Keeps 1nt PAS coordinates so can do matching as before, only difference is score is relative usage instead of TPM. 
2. Output file to report per 'region' (terminal exon, gene) relative usage. Applies to tools like LABRAT & APAlyzer that report a single relative usage metric for a whole region and not each poly(A) site. Coordinates may not be required down the line for summary workflows but am keeping as a BED to maintain consistency.

README updated, example files also added. Feedback much appreciated but let's get this done asap so we can start updating the execution workflows!!!


execution_workflows/execution_output_specification.md
```diff
@@ -10,6 +10,8 @@ Outputs depend on features available for the participant (i.e. not all participa
   - BED file with identified poly(A) sites with single nucleotide resolution
 - quantification benchmarking event:
   - BED file with identified unique poly(A) sites and TPM values for each site
+  - BED file with identified unique poly(A) sites and relative usage values for each site
+  - BED file with coordinates of regions (e.g. terminal exons, genes) and the relative usage values for each region
```
ninsch3000, line 14: I'm afraid I don't quite understand what the "relative usage value" is for a region. Relative usage of distal site? Relative usage of the region compared to other regions? Can you elaborate a bit, also below in section "Format 05"?

SamBryce-Smith:
Thanks for comments @dominikburri & @ninsch3000! Sorry for delayed response again. Will respond to Dominik's  comments below :)

RE regions for Format 05, just to be clear I don't view it as essential at all. My main motivation was that I don't think we're fully clear on how we plan to implement downstream processing for these metrics in the summary workflows. I thought it would be safer to retain too much info from the Ewfs now so we can make the updates to Ewfs first, then make the right decision about implementing later without going back on ourselves.

> How is the region you mention obtained? Is it from the tool directly? Or from the gencode annotation GTF? (do we actually need that for each execution workflow? based on [this](https://github.com/iRNA-COSI/APAeval/tree/main/execution_workflows) probably not.)

1 - Either from the output tables (e.g. DaPars(2), APAtrap report the last exons for sure from top of my head) or the GTF as a backup. Do we need the region? Quite possibly not, perhaps the gene ID / transcript ID would be enough for us and we can go back to the reference GTF in the summary workflow.

> If the region considered is from the tool, how do we make it comparable? Is that part of the summary workflow?

> If we define the regions from the GTF, do we disregard or try to match the ones from the tools (if those report that)?

My thinking at the time was to do it as part of the summary workflow. In pretty much all the cases I can think of the 'region' considered by the tool comes from the annotation anyway (i.e. searching last exons for internal polyA). Would then subset the ground truth for polyA sites overlapping this region.

RE second point, it's a fair one I'm not actually sure how many tools qualifying for this metric actually report/try to identify a downstream polyA site/extended terminal exon.

> Would each tool report as many BED outputs as possible?

Yes that was my thinking! If it doesn't report anything qualifying for format 04 or 05 we don't generate those outputs in the Ewf. 

> That is, if a tool reports per site PAS, it could also report per-PAS relative usage and per-region PAS usage, given we use the GTF to obtain regions of interest.

I think a tool like PAQR could do this if we implemented the 'weighted_avg_exon_lengths` rule. I.e. per PAS we report the fractional usage, for the terminal exon we'd report the score from weighted avg exon lengths. PAQR's annotation start with extracting terminal exons from the GTF from memory. The per-region PAS usage/score is only generated if the tool specifically generates a metric fitting this description (e.g. LABRAT, APAlyzer)

> In more detail, would the most general tool output BED format 1 and 2 as before, but also BED format 4 and 5 computed by us? And the most restrictive tool (in the sense of per-site and region PAS reporting) only report BED format 5.

To my mind yes, 1 + 2 + 4 / 1 + 4 would be the most common combinations. Format 5 is really only for a few specific tools.

**Perhaps to make things easier we could just make a TSV of gene_id/transcript_id | score for format 05 (gene ID for 'gene' level tools e.g. LABRAT, transcript id for 'terminal exon' level tools). Would be a bit wary of having different ID types in the same format though...**

---

After meeting 13/07 any references to format 05 are irrelevant, but I'm keeping comments up for future discussion...

ninsch3000 (approved):
Thanks a lot @SamBryce-Smith!

dominikburri (approved):
Hi @SamBryce-Smith, thank you for the clarifications!
I like your idea of regions as gene or transcript IDs, but I agree it could get messy.
I'm not completely sure why you now removed BED format 05 - relative usage per region. Would this not lead to some tools be dropped, because they only provide this kind of information? 

Maybe I missed some discussions, so I don't want to hold it back from here. The spec BED 04 looks fine.

PR #385, opened 2022-07-18
execution workflow: DaPars2 - output  Format 04 fractional relative usage quantification BED file

### LINKED ISSUE #383: Execution workflow: DaPars2 - report Format 04 BED file in workflow (per-PAS fractional relative usage)
Parent issue - #382 

Per updated execution workflow [output specifications](https://github.com/iRNA-COSI/APAeval/blob/main/execution_workflows/execution_output_specification.md), We need DaPars2 to report the per-PAS fractional relative usage in a format 04 BED file.

For each input transcript model, DaPars2 reports a 'PDUI' (% distal usage index), representing the fraction/percent the distal site is expressed relative to the total expression of proximal and distal site of that locus. Since DaPars2 uses a 2-site model, we can also report the usage of the proximal site by 1 - PDUI.

SamBryce-Smith:
Fixes #383

## Checklist

- [x] I have tested my code

Successful run with test data 
dapars2-test-run-Screenshot from 2022-07-18 14-11-28

The 'Format 04' BED file fits the specifications. Since DaPars2's PDUI is the fraction the distal site is used relative to total expression of the region and DaPars uses a '2 site model' (i.e. only 2 sites predicted/reported per-region), we can calculate the relative usage of the proximal site by 1 - PDUI for region. For the distal site we just report the calculated PDUI.

Note that DaPars2's PDUI seems to only ever be reported to 2 d.p., so for sake of consistency I opted to also only report the proximal polyA site fractional usage to 2 d.p.
dapars2-format-04-bed-Screenshot from 2022-07-18 14-17-00

---

**Unrelated to this PR but generally to the design of the execution workflow**, it looks like we missed DaPars2 on our last round of updating. I've noticed the following are out of sync with our templates/Ewf specifications:
- This workflow **does not have 'challenge output flags'** i.e. it always outputs identification & relative usage quantification outputs. For sake of time I didn't bother to change this now as I felt getting example format 04 output was more important, but if considered pressing I can do this/implement in a separate PR.
- The workflow sits in `execution_workflows/DaPars2/Dapars2_snakemake` when it should probably just sit underneath `execution_workflows/DaPars2`. Not a huge deal but a little inconsistent


dominikburri (approved):
Hi @SamBryce-Smith, many thanks for the work!
The code looks fine and many thanks for the detailed descriptions. 
I ran the provided test without problems and the resulting `04.bed` files look fine and can be displayed in IGV. 
You mention that the workflow does not follow the latest standard. For me, this is is not a problem, maybe we can collect those in a common issue though? At least that we are aware and if we have resources that we can work on it. 
Anyway, looks good to me.

I have some questions though (not really related to this PR I fear):
- The test is with 2 genes, we get though 12 proximal and distal sites for Wnk1 and 3 for Cxadr. How would one compare them to other relative usages from other tools? By transcript ID? 
- I glimpsed at IGV and it seems it does not follow the annotated transcripts very closely. I know this is the output from the tool but it raises the question on how we would select appropriate sites for comparison.
Screenshot from 2022-07-25 15-29-50
Note that some sites are overlapped in this condensed view.

yuukiiwa (approved):
Thanks, Sam!! Looks good to me!!

PR #393, opened 2022-07-25
Execution workflow: IsoSCM - report Format 04 BED file in workflow (per-PAS fractional relative usage)

### LINKED ISSUE #389: Execution workflow: IsoSCM - report Format 04 BED file in workflow (per-PAS fractional relative usage)
Parent issue - https://github.com/iRNA-COSI/APAeval/issues/382

Per updated execution workflow [output specifications](https://github.com/iRNA-COSI/APAeval/blob/main/execution_workflows/execution_output_specification.md), We need IsoSCM to report the per-PAS fractional relative usage in a format 04 BED file.

According to https://github.com/shenkers/isoscm, the `compare` step outputs a relative usage in the `site_usage` column. This *should* be equivalent to a fractional polyA site usage (quoting from the documentation/README):

> We estimate relative usage of a polyadenylation site by considering the read density in the segments up and downstream of a change-point, such that
usage = 1 - ( covdn / covup )

From memory we don't implement the compare step in our workflow yet, so this one requires a little more work (would prioritise the others first).

faricazjj:
Fixes #389 

IsoSCM now has two parameters: run_identifcation and run_relative_usage_quantification which are booleans indicating which challenge(s) we want outputs for.

With run_identification set to true, an example of the output file is:
Screen Shot 2022-07-26 at 3 03 03 PM

With run_relative_usage_quantification set to true, an example of the output file is:
Screen Shot 2022-07-26 at 3 01 03 PM

and the new relative usage quantification workflow is:
Screen Shot 2022-07-26 at 3 00 42 PM

Setting both run_identification and run_relative_usage_quantification to true, we get the following workflow:
Screen Shot 2022-07-26 at 3 02 42 PM

The output files for both the challenges are:
Screen Shot 2022-07-26 at 3 02 54 PM

## Checklist

- [x] I have tested my code


faricazjj:
Hi @SamBryce-Smith! I'm reading [IsoSCM github readme ](https://github.com/shenkers/isoscm) where currently, we get results for identification challenge from one of the output files from the assemble step. This file seems to contain the identified changepoints.
Screen Shot 2022-07-26 at 11 32 38 AM
However, the compare command "reports the differential usage of each identified change-point", which I expected to show the site usages of the same sites above. But I see less and also different sites in the compare step output file
Screen Shot 2022-07-26 at 11 35 09 AM

Are we supposed to have the same exact sites for identification and quantification, and then only difference is the score column? Or do these two files look correct?

execution_workflows/IsoSCM/README.md
```diff
@@ -68,9 +68,17 @@ Parameters relevant to the workflow itself are:
 - `star_genome_index` - absolute path to the star genome index folder
 
 ### Running the identification workflow
+- Set 'run_identification' parameter in conf/modules.config to 'true'
```
yuukiiwa, line 71:
```
- Set `run_identification` parameter in conf/modules.config to `true`
```

execution_workflows/IsoSCM/README.md
```diff
@@ -68,9 +68,17 @@ Parameters relevant to the workflow itself are:
...
 - Change 'identification_out_suffix' parameter in conf/modules.config to the desired file name that ends with '.bed'
```
yuukiiwa, line 72:
```
- Change `identification_out_suffix` parameter in conf/modules.config to the desired file name that ends with `.bed`
```

execution_workflows/IsoSCM/README.md
```diff
@@ -68,9 +68,17 @@ Parameters relevant to the workflow itself are:
...
 - Change the 'output_dir' to the desired directory name under IsoSCM/results/isoscm the identification challenge output will be in  
-- An example sample sheet is samplesheet_example_files_identification.csv.
+- An example sample sheet is samplesheet_example_files.csv.
+- Run the pilot benchmark nextflow pipeline with nextflow main.nf --input samplesheet_example_files.csv
+
+### Running the relative usage quantification workflow
+- Set 'run_relative_usage_quantification' parameter in conf/modules.config to 'true'
```
yuukiiwa, line 78:
```
- Set `run_relative_usage_quantification` parameter in conf/modules.config to `true`
```

execution_workflows/IsoSCM/README.md
```diff
@@ -68,9 +68,17 @@ Parameters relevant to the workflow itself are:
...
+- Change 'relative_usage_quantifciation_out_suffix' parameter in conf/modules.config to the desired file name that ends with '.bed'
```
yuukiiwa, line 79:
```
- Change `relative_usage_quantifciation_out_suffix` parameter in conf/modules.config to the desired file name that ends with `.bed`
```

execution_workflows/IsoSCM/README.md
```diff
@@ -68,9 +68,17 @@ Parameters relevant to the workflow itself are:
...
+- Change the 'output_dir' to the desired directory name under IsoSCM/results/isoscm the relative usage quantification challenge output will be in
```
yuukiiwa, line 80:
```
- Change the `output_dir` to the desired directory name under IsoSCM/results/isoscm the relative usage quantification challenge output will be in  
```

yuukiiwa (approved):
Thanks so much, @faricazjj!! Looks good to me!! I only have very minor comments on the `README.md`

execution_workflows/IsoSCM/bin/check_input_params.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def main(args=None):
+    args = parse_args(args)
+    if args.RUN_IDENTIFICATION == True:
+        if not args.IDENTIFICATION_OUT_SUFFIX.endswith(".bed"):
+            msg = "The identification output file name should end with '.bed'"
+            sys.exit(msg) 
+    if args.RUN_RELATIVE_USAGE_QUANTIFICATION == True:
+        if not args.IDENTIFICATION_OUT_SUFFIX.endswith(".bed"):
```
dominikburri, line 28: should it be `RELATIVE_USAGE_QUANTIFICATION_OUT_SUFFIX`?

execution_workflows/IsoSCM/bin/postprocess.py
```diff
@@ -0,0 +1,71 @@
...
+            # since the two samples for IsoSCM compare are the same, the site usage values
+            # are the same (e.g. 0.9895184913950934,0.9895184913950934), so we can arbitrarily take the first 
+            score = row[11].split(",")[0]
+            output = [chrom, str(start), str(end), name, score, strand]
+            outputs.append(output)
+
+    return outputs
+
+
+def get_identification_output(outputs, file_out):
```
dominikburri, line 46: maybe rename to `write_identification_output`?

execution_workflows/IsoSCM/modules/isoscm_compare.nf
```diff
@@ -0,0 +1,28 @@
+// Import generic module functions
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+params.options = [:]
+def options    = initOptions(params.options)
+
+process ISOSCM_COMPARE {
```
dominikburri, line 7: Can you add some description to this step as in the assemble step? 
Is this where two samples/conditions are needed and we the same?

dominikburri (commented):
Hi @faricazjj,
thank you for the work! 
I ran the test files without problems. I had to set `run_star_genome_generate = true` because I deleted the whole results directory. 

The test contains 2 genes, Cxadr on the positive strand and Wnk1 on the negative strand. 

However, I get for Cxadr entries for the positive and negative strand, see below. The output you report is the same. It is elusive to me how IsoSCM can report a negative strand PAS in the control_r1 sample when the indexing etc is done with the 2 gene gtf file. I don't know how IsoSCM works, maybe I miss something. 
Screenshot from 2022-08-02 15-29-21

As for Wnk1, the srsf3_r1 sample reports two PAS in a terminal exon:
```tsv
chr6	119923962	119923963	chr6|119923962:119923963|-	0.9674360883593944	-
chr6	119925602	119925603	chr6|119925602:119925603|-	0.5016820857863753	-
```
How are the numbers interpreted? should the two values be added up to 1? or is it something like the cumulative fraction of read coverage (i.e. the proximal peak - negative strand - covers 50% and the distal peak 97% of the reads)?

Given @SamBryce-Smith discussion on relative usage, would it be possible to report the transcript ID in the name column of the output BED file? This would help in tracking the membership of PAS and subsequent comparison?

faricazjj:
> Thanks so much, @faricazjj!! Looks good to me!! I only have very minor comments on `README.md`

Thanks @yuukiiwa ! :D I've addressed your comments in the readme

faricazjj:
> Hi @faricazjj, thank you for the work! I ran the test files without problems. I had to set `run_star_genome_generate = true` because I deleted the whole results directory.
> 
> The test contains 2 genes, Cxadr on the positive strand and Wnk1 on the negative strand.
> 
> However, I get for Cxadr entries for the positive and negative strand, see below. The output you report is the same. It is elusive to me how IsoSCM can report a negative strand PAS in the control_r1 sample when the indexing etc is done with the 2 gene gtf file. I don't know how IsoSCM works, maybe I miss something. Screenshot from 2022-08-02 15-29-21
> 
> As for Wnk1, the srsf3_r1 sample reports two PAS in a terminal exon:
> 
> ```generic-db
> chr6	119923962	119923963	chr6|119923962:119923963|-	0.9674360883593944	-
> chr6	119925602	119925603	chr6|119925602:119925603|-	0.5016820857863753	-
> ```
> 
> How are the numbers interpreted? should the two values be added up to 1? or is it something like the cumulative fraction of read coverage (i.e. the proximal peak - negative strand - covers 50% and the distal peak 97% of the reads)?
> 
> Given @SamBryce-Smith discussion on relative usage, would it be possible to report the transcript ID in the name column of the output BED file? This would help in tracking the membership of PAS and subsequent comparison?

Thanks @dominikburri ! :D I've addressed your comments in the code. 

As for the first question about the discrepancies in the strands, that's an excellent point, thanks so much for looking into it! I'm not very familiar with star alignment for XS tag, it's my first time using it. I'm also not too sure about strandedness and single/paired end reads. But currently, the star alignment implementation in the workflow uses --outSAMstrandFiled intronMotif as suggested by Chilam who worked on the first draft of the workflow. I read that this is for unstranded reads. Could this be messing up with the strands in the final output? I'm checking with @mrgazzara about the strandedness of our data. But other than the star alignment step, I can't find anything else that could possibly mess up with the strands in the final output.

As for the second question, from what I understand it seems like the numbers wouldn't add up to one according to the formula that they use. Sam wrote that according to https://github.com/shenkers/isoscm, the `compare` step outputs a relative usage in the `site_usage` column. This *should* be equivalent to a fractional polyA site usage (quoting from the documentation/README):

> We estimate relative usage of a polyadenylation site by considering the read density in the segments up and downstream of a change-point, such that
usage = 1 - ( covdn / covup )

That's what I understand from the formula above and it's more like cumulative fraction of the read coverage. @mrgazzara lmk if this is the right interpretation!

As for the third question, they don't provide transcript id in the output that looks like this
Screen Shot 2022-08-04 at 3 45 05 PM
So there's nothing we can directly extract transcript ids from unless we add an extra step to extract it from the annotation file :(

dominikburri:
Hi @faricazjj, thank you for the updates!

1. Do you have more info about the strands? is it from the XS tag? That is, is it solved?
2. Do you think you can report relative usage within the 3'UTR/TE instead of the cumulative fraction? I think then it would be more comparable to other tools. I was under the impression that the relative usage should sum to 1 in a given region (usually terminal exon). 
3. Don't worry about the transcript IDs. As you say, we probably extract it from the annotation file. I just thought, if it is readily available we could've use it, but no problem if not.

faricazjj:
@dominikburri thanks for following up! 

1. I discussed with @mrgazzara on the strands and it could be because the test data is unstranded. I checked IsoSCM code and they deal with unstranded data by inferring strands from spliced reads, I assume using the XS tag. I didn't really understand what XS tags are for and when looking into it, I found a similar question to ours. https://github.com/alexdobin/STAR/issues/567. That's why I think it's not following the strands from the annotation file.
Here is the IsoSCM code in IsoSCM.java that writes the strands to the output file. It suggests that they get the strands by inferring strands from spliced reads, which is what the XS tag is described to do. 
Screen Shot 2022-08-08 at 4 59 25 PM

2. For this I'm actually not sure. @SamBryce-Smith @mrgazzara did you guys look into how the raw IsoSCm's site usage is comparable to other tools' relative usage in your POC?

faricazjj:
Update: @mrgazzara and I have been discussing about how to investigate the strand issue further. Right now, I've generated two sets of outputs, one from our unstranded test data and the other one from the stranded mmus cortex adult R1 stranded data. Matt is going to help to see if the strandedness of the data is causing the strand issue pointed out above by @dominikburri

mrgazzara:
I've done a lot of digging through the issue that @dominikburri brings up and I think that everything that IsoSCM is finding, even in the test data, is ok / expected. I'll address the points in your most recent comment:

**1) The strand issue is ok!** 
The subsetted bam files were made by taking the boundaries of the gene annotation for those two genes, adding a few kb, and dumping all those reads that intersect that region to a bam file. In the case of Cxadr there is an antisense gene, Btg3, in this region. Everything that IsoCSM reports seems to be somewhat annotation agnostic...So I think just because the end of Btg3 is near Cxadr we are also getting quantification for that region as well. See the screen shot below showing also the regions that IsoSCM finds as having APA:

hgt_genome_161a3_5f72e0

As you can see the upstream and downstream segments the algorithm considers for Cxadr (left, + strand) and Btg3 (right, - strand) look reasonable and are on the proper strand. 

**2) Relative usage**
We were not handling things properly before. When you tested this ONLY the proximal PAS was being reported (what IsoSCM calls the `changepoint` and we reported the `site_usage`, which is equivalent to the percent of proximal PAS used. IsoSCM behaves similar to DaPars in that it deals only with pairs of PAS at a time. In some genes with >2 PAS it seems to slide a window along and quantifies groups of adjacent PAS together. So with the edits I discussed with @faricazjj we will now also report the dPAS coordinates and can report the relative usage of the dPAS as `1 - site_usage`. Pairs of PAS considered in each line will therefore sum to one. The example you showed above where one PAS is ~97% and the other is ~50% I assume is from one such case of >2 PAS in the terminal exon. In the updated version there should be 4 PAS (before we were only showing the 2 proximal PAS and quantifications). 

**3) Transcript names**
Unfortunately the output is not associated / marked with any gene or transcript information. The relative quantification SWF will do a TE matching to these output to assign them to a gene. What we will need is to know which pairs of PAS are derived from teh same line in the file. I've talked to @faricazjj and this will be doable / easy to implement.

faricazjj:
@mrgazzara @dominikburri thank you so much for reviewing and commenting!! :D I've just incorporated the name change solution from Matt and merged the bug fix PR here. The output relative quant output for control_r1 in our test data now looks like this:
Screen Shot 2022-09-30 at 10 49 25 AM

It's ready for rereview!

dominikburri (approved):
Thank you @faricazjj, I ran the test with singularity without problems. 
The relative output BED files look fine, the proximal and distal sites add up to 1 and the sites are identical to the identification sites. 
I only have some minor comments, nothing stopping the approval here.

mrgazzara (approved):
@faricazjj Looks good to me!

mrgazzara (approved):

PR #396, opened 2022-08-02
[Bug fix] Dapars: Skip 'chr' prefix check for non main chromosomes

### LINKED ISSUE #395: Dapars errors out if there is a contig in the sample file
We added a check to make sure the input files have 'chr' prefixes for all chromosomes. This was done to prevent Dapars from erroring out with a not so obvious error message when NO row contains 'chr' prefix. 

However, there are contigs in input files that by default contain to 'chr' prefix which we should allow. These row will be ignored by Dapars and we shouldn't be preventing the execution workflow from continuing by throwing an error when we encounter these rows. 

Screen Shot 2022-08-01 at 8 23 39 PM

Our check should allow files containing non main chromosomes to pass without 'chr' prefix. These rows will be skipped by dapars by default. Our check should only make sure at least one row to have 'chr' prefix, as required by Dapars to not throw an error.

faricazjj:
Fixes #395 

We added a check to make sure the input files have 'chr' prefixes for all chromosomes. This was done to prevent Dapars from erroring out with a not so obvious error message when NO row contains 'chr' prefix. 

However, there are contigs in input files that by default contain to 'chr' prefix which we should allow. These row will be ignored by Dapars and we shouldn't be preventing the execution workflow from continuing by throwing an error when we encounter these rows. 

Screen Shot 2022-08-01 at 8 23 39 PM

Our check should allow files containing non main chromosomes to pass without 'chr' prefix. These rows will be skipped by dapars by default. Our check should only make sure at least one row to have 'chr' prefix, as required by Dapars to not throw an error.

## Checklist

- [x] I have tested my code

### DIFF
--- a/execution_workflows/DaPars/bin/check_bedgraph.py
+++ b/execution_workflows/DaPars/bin/check_bedgraph.py
@@ -4,33 +4,40 @@ import sys
 import argparse
 
 def parse_args(args=None):
-	Description = "Check DaPars bedgraph input to have leading chr for all sequence regions in main chromosomes (1-22,X,Y,M). Otherwise, throw an error."
-	Epilog = "Example usage: python check_bedgraph.py <FILE_IN>"
+        Description = "Check DaPars bedgraph input to have leading chr for at least one sequence region. Otherwise, throw an error."
+        Epilog = "Example usage: python check_bedgraph.py <FILE_IN>"
 
-	parser = argparse.ArgumentParser(description=Description, epilog=Epilog)
-	parser.add_argument("FILE_IN", help="Input bedgraph file.")
-	return parser.parse_args(args)
+        parser = argparse.ArgumentParser(description=Description, epilog=Epilog)
+        parser.add_argument("FILE_IN", help="Input bedgraph file.")
+        return parser.parse_args(args)
 
 def check_bedgraph(file_in):
-	"""
-	This function checks that there is leading chr for all sequence regions
-	Otherwise, throw an error
-	:param file_in: bedgraph file to be checked
-	:return: N/A
-	"""
-	fin = open(file_in, "rt")
+        """
+        This function checks that there is leading chr for at least one sequence region
+        Otherwise, throw an error
+        :param file_in: bedgraph file to be checked
+        :return: N/A
+        """
+        fin = open(file_in, "rt")
+        num_no_chr = 0
+        num_line = 0
+        for line in fin:
+            num_line += 1
+            if line[0] != '#' and line[:3] != "chr":
+                num_no_chr += 1
 
-	for line in fin:
-		if line[0] != '#' and '.' not in line and line[:3] != "chr":
-			msg = "Found a row in the sample file " + file_in + " without leading 'chr' in the chromosome column for a main chromosome (1-22, X, Y, M).\n" + \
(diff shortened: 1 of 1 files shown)


execution_workflows/DaPars/bin/check_bedgraph.py
```diff
@@ -21,9 +21,9 @@ def check_bedgraph(file_in):
 	fin = open(file_in, "rt")
 
 	for line in fin:
-		if line[0] != '#' and line[:3] != "chr":
-			msg = "Found a row in the sample file " + file_in + " without leading 'chr' in the chromosome column.\n" + \
-				" Please use a file with leading 'chr' for all sequence regions."
+		if line[0] != '#' and '.' not in line and line[:3] != "chr":
```
dominikburri, line 24: As I understand it, if no dot is in the line, nor a comment (#) and it does not start with "chr", it is not a valid format.
I don't think the dot check will help, as it can be anywhere in the line. In the example you give, the column "rep" contains dots in the decimals for example. 
Maybe you need to check whether the line starts with `(1-22,X,Y,M)`, as you describe in the message.

dominikburri (commented):
Hi @faricazjj, 
thanks for the fix, see my comment. 
I didn't test it but figured you tested it on the current genome (where it supposedly works). 
Did you test it with a file that triggers the if statement, i.e. chromosomes without preceding "chr"?

execution_workflows/DaPars/bin/check_bedgraph.py
mrgazzara: This is a good point. Maybe another, more general, fix is to check that there are _any_ lines in the bedgraph that lead with "chr", otherwise throw the error message. In theory, someone may use this execution workflow for non-human organisms so checking for the specific chromosome nomenclature we see in human (or mouse) might be limiting?

dominikburri: hm yeah that would be nice. Though I'm not sure if DaPars needs a specific nomenclature because the gtf is hardcoded or so. If not, then the best would be to check whether the sample and gtf have same chromosome names. I think that's what we do in the validation within the quantification swf: `summary_workflows/quantification/quantification_dockers/q_validation/validation.py`

faricazjj: Oh good point, thanks guys! We need at least one row with leading chr otherwise dapars will error out in the code with a not so obvious error that's it's caused by no leading chr, and that's why we added this check to give a straightforward error message in case none of the rows in the chromosome column have leading chr. I wasn't sure how other non human organisms chromosome nomenclature looks like so I'll change the check to make sure there is at least one row with chr since that's what dapars needs to run without a cryptic error

faricazjj: I'm rerunning dapars with the new check I mentioned above, I'll update this PR once it passes the run just to double check! @dominikburri @mrgazzara

mrgazzara (approved):
This looks good to me 👍

dominikburri (approved):
Hi @faricazjj, 
thank you for the fix! Looks good to me :) 
Though I'm not sure why indentation increased (from 4 to 8?). But if the code runs I don't see a problem.

PR #403, opened 2022-08-08
[Bug Fix] APAtrap cast output to string before parsing and avoid time out error

### LINKED ISSUE #402: [Bug] APAtrap error when splitting int object and time out error
1. When converting APAtrap output to absolute TPM quantification challenge output bed file, a column containing a list of TPMs is taken and parsed by splitting by comma. This column is known to contain strings and so is directly split by comma, which has worked for most of the dataset. 

However, while running APAtrap on one of Keratinocyte samples, one row is written as int in APAtrap output file while the rest were string as expected. Hence, this row that is written as int caused the workflow to error out with the following message:
Screen Shot 2022-08-03 at 9 04 40 PM

A solution for this is to always cast the rows to string before splitting as such:

Screen Shot 2022-08-08 at 2 49 42 PM

2. APAtrap takes a long time when processing real datasets. Even after processing the samples per chromosome, it could take at least 12 hours depending on the number of samples in the sample sheet. As such, APAtrap is prone to time out errors or out of memory errors. One solution is to add label 'process_high' in APAtrap processes to give it more time to run.

faricazjj:
Fixes #402 

This PR solves two errors encountered when running APAtrap with real data.

1. When converting APAtrap output to absolute TPM quantification challenge output bed file, a column containing a list of TPMs is taken and parsed by splitting by comma. This column is known to contain strings and so is directly split by comma, which has worked for most of the datasets. 

However, while running APAtrap on one of the Keratinocyte samples, one row was written as int in the APAtrap output file while the rest were written as string as expected. Hence, this row caused the workflow to error out with the following message:
Screen Shot 2022-08-03 at 9 04 40 PM

A solution for this is to always cast the rows to string before splitting as such:

Screen Shot 2022-08-08 at 2 49 42 PM

2. APAtrap takes a long time when processing real datasets. Even when processing the samples per chromosome, it could take at least 12 hours depending on the number of samples in the sample sheet. As such, APAtrap is prone to time out errors or out of memory errors. One solution is to add the label 'process_high' in APAtrap processes to give it more time to finish running to avoid time out errors.

## Checklist

- [x] I have tested my code

### DIFF
--- a/execution_workflows/APAtrap/README.md
+++ b/execution_workflows/APAtrap/README.md
@@ -13,7 +13,7 @@ to create the nextflow pipeline flow of this module
 
 ## Running APAtrap workflow
 
-## Input & pre-processing
+### Input & pre-processing
 Required files are to be specified in the input `samplesheet_example_files.csv`. Each row in the sample sheet has four
 columns:
 
@@ -34,13 +34,13 @@ input file you are using the absolute path to the file, with the four column nam
 names in `samplesheet_example_files.csv`.
 
 ### Running with Docker or Singularity
-## Docker
+#### Docker
 This workflow uses docker containers. To run with docker, make sure that docker is installed and running 
 (e.g. to ensure docker is running, run the command `docker --help` and a help message should be printed).
 Additionally, make sure that line 49 in Apatrap/nextflow.config file `docker.enabled=true` is uncommented while line
 51 `singularity.enabled=true` is commented out.
 
-## Singularity
+#### Singularity
 To run with singularity, comment out line 49 in Apatrap/nextflow.config file `docker.enabled=true` and make sure that line
 51 `singularity.enabled=true` is uncommented.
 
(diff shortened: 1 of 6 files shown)


yuukiiwa (approved):
Thanks, Farica! Looks good to me!!

dominikburri (commented):
Hi @faricazjj, many thanks for the fixes! 
I successfully tested it with the provided test data and am quite sure your fixes work for the real data you mention.

I'm kind of nitpicking now.  The _process_high_ label uses by default 12 cpus, but the individual processes use only 1 cpu as I gauge from the code. Maybe you can use a label like this:
```
withLabel:process_big_mem {
    cpus   = { check_max( 1 * task.attempt, 'cpus' ) }
    memory = { check_max( 128.GB * task.attempt, 'memory' ) }
    time   = { check_max( 72.h * task.attempt, 'time' ) }
  }
```
Also, not sure whether all 3 processes need the _process_high_ label. Maybe only one process is doing the heavy computation?
I'm just trying to tailor the execution, such that resources are sensibly used.

grexor (dismissed):
looks good

faricazjj:
> Hi @faricazjj, many thanks for the fixes! I successfully tested it with the provided test data and am quite sure your fixes work for the real data you mention.
> 
> I'm kind of nitpicking now. The _process_high_ label uses by default 12 cpus, but the individual processes use only 1 cpu as I gauge from the code. Maybe you can use a label like this:
> 
> ```
> withLabel:process_big_mem {
>     cpus   = { check_max( 1 * task.attempt, 'cpus' ) }
>     memory = { check_max( 128.GB * task.attempt, 'memory' ) }
>     time   = { check_max( 72.h * task.attempt, 'time' ) }
>   }
> ```
> 
> Also, not sure whether all 3 processes need the _process_high_ label. Maybe only one process is doing the heavy computation? I'm just trying to tailor the execution, such that resources are sensibly used.

@dominikburri thank you for the suggestion! I initially thought these are the cap for the resources but after seeing your comment and also [nextflow documentation](https://nf-co.re/usage/configuration) these are actually allocated resources. Thanks for being meticulous! :D I've updated the PR as suggested and added in the README a note on it!

dominikburri (approved):
Thank you @faricazjj! LGTM :)

yuukiiwa (commented):
Thanks, @faricazjj!! Looks good to me!!

yuukiiwa (approved):

PR #405, opened 2022-08-09
SWF: modules for code deduplication

### LINKED ISSUE #374: Summary workflows: Control code duplication
## Problem
We have 3 distinct benchmarking events, each of which has its own summary (=benchmarking) workflow. These [summary workflows](https://github.com/iRNA-COSI/APAeval/tree/main/summary_workflows) are very similar to each other and only differ in a couple of functions. Each workflow consists of 3 steps, each of which is run in a separate docker container. The containers are built separately for each of the 3 steps for each of the 3 events, from Dockerfiles residing in a directory structure as described below. The 3 events do share some functions though, which we need to keep synchronized across events. In the current implementation, we keep copy-pasting the functions between events whenever they change, which will inevitably cause inconsistencies at some point.
```
summary_workflows
    |-- event_A
        |-- step_1
            |-- Dockerfile
            |-- requirements.txt
            |-- script_1A
            |-- module_X
                 |-- function_X
        |-- step_2
            |-- Dockerfile
            |-- requirements.txt
            |-- script_2A
            |-- module_Y
                 |-- function_Y
        |-- step_3
          .....
    |-- event_B
        |-- step_1
            |-- Dockerfile
            |-- requirements.txt
            |-- script_1B
            |-- module_X
                 |-- function_X
        |-- step_2
            |-- Dockerfile
            |-- requirements.txt
            |-- script_2B
            |-- module_Y
                 |-- function_Y
        |-- step_3
          .....
    |-- event_C
          .....
```
While the basic structures of script_1A and script_1B are similar, they do have event specific adaptations, so I suppose we can't do much better than duplicating the code. However, module_X really is the same in both events and we want to keep it synchronized.     
If we understand correctly, docker requires files to be included in a container to reside inside the _build context_ 

> (see [here](https://docs.docker.com/engine/reference/builder/#add), "The <src> path must be inside the context of the build; you cannot ADD ../something /something, because the first step of a docker build is to send the context directory (and subdirectories) to the docker daemon.")   

so that's why we haven't moved `module_X` to the `summary_workflows` directory level.     

### _How can we keep `module_X` synchronised between events?_   
And: is there really nothing we can do for `script_1A` and `script_1B`?

## Possible solutions?
1. Circumvent the above described docker requirement?
  Would [the single file archive suggested in the first answer here](https://stackoverflow.com/questions/37789984/how-to-copy-folders-to-docker-image-from-dockerfile) work/make sense?
2. Add `module_X` to PyPi and `requirements.txt`?
   How much work is this, especially in case we need to update `module_X` frequently?
3. Use [Git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules)?
   How much work is this and how robust is it? Can the containers be built smoothly with that approach?
4. Do something better?
   What?

ninsch3000:
Fixes #374    
Functions for calculating metrics and creating OEB compliant JSON files have been moved from the individual "dockers" directories to the APAeval utils and summary workflow directories, respectively. With this setup the functions only have to be edited in one place and can from there be installed in different docker containers for different summary workflows. Thanks to @pjewell-biociphers for implementation

## Checklist

- [ ] I have tested my code


ninsch3000:
To the reviewers: testing as is will not work because I was too eager and already changed the branch in the requirements.txt to main. So, for testing, change back to "code-deduplication-modules" , in requirements of q_validation and q_metrics

dominikburri (approved):
Thank you @ninsch3000, 
I added a commit with instructions for locally building the two modules. In this way, I was able to locally debug the code. 
After the modification you mention in a previous comment, I successfully ran the provided tests: `test.pd.mm10_test.bed` with challenge `test.gt.mm10_test` and `embryonicCortex.PAPERCLIP.mm10.bed` with challenge `Qx.mm10`. 
I could not check whether the empty dirs (JSON_templates) were deleted as I still have pycaches in there. 
Anyway, LGTM.

PR #430, opened 2022-09-23
feat: unify TP calculation in aQ swf

### LINKED ISSUE #425: SWF: unify true positive calculation
Currently there are differences in TP calculation between SWFs
- aQ: TP = matched.loc[matched["score_p"] > 0].shape[0]
- rQ: TP=len(matched)
- I: TP = matched[~matched.duplicated(GT_cols)].shape[0]  

We agreed in the meeting to unify the calculation between the SWFs; use the one from I ("matched" without duplicated GT sites)

dominikburri:
Fixes #425 for the absolute Quantification challenge.

## Notes
Changes in True positives
* The test files give the same values in _sensitivity_ , _precision_ and _jaccard index_
* For a real example, I used an older result of _Mayr_CD5B_R3_ on PAQR, I obtain more TP now. I guess this is because the TPM expression in the prediction can be > 0.

## Checklist

- [x] I have tested my code


yuukiiwa (approved):
Thanks so much, @dominikburri!! This looks good to me!

faricazjj (approved):
Thank you so much @dominikburri LGTM!!

PR #431, opened 2022-09-27
[Bug Fix] IsoSCM identification output should be from assemble step

### LINKED ISSUE #429: [Bug Fix] IsoSCM identification output should be from compare step
Currently, we get results for identification challenge from one of the output files from the _assemble_ step. Initially, this file seems to contain the identified changepoints just by looking at the content of the file.
Screen Shot 2022-07-26 at 11 32 38 AM

But as I read further beyond the assemble step, the second step they describe is the _compare_ command. The _compare_ command "reports the differential usage of each identified change-point", which I expected to show the site usages of the same sites as the sites from the previous _assemble_ step shown in the screenshot above. But I see less and also different sites in the _compare_ step output file
Screen Shot 2022-07-26 at 11 35 09 AM

This led me to look more into whether we should be extracting sites for identification challenge from the _assemble_ or _compare_ step.

As additional context, the steps to run isoscm are:
1. run _assemble_ step which creates a tmp folder: isoscm/tmp/{sample}.cp.filtered.gtf
2. run _compare_ step, this requires the xml file output from assemble step for two samples, but since we're getting site usage per sample, I put the same sample twice as input to obtain the following output:
Screen Shot 2022-07-26 at 12 18 54 PM

Even though the first isoscm/tmp/{sample}.cp.filtered.gtf file contains changepoint locations, I'm not entirely sure we should obtain identification output sites from there since it's in a tmp folder and the github readme doesn't explain what the files in the tmp folders are--they only explained the files outside of the tmp folder from _assemble_ step i.e they explained the files from step 2 above but not files from step 1. I think the sites for identification might have to be obtained from the _compare_ step output (i.e. isoscm/compare/{sample}.txt). After reading their readme, I checked their paper and saw that in their paper, they didn't mention _assemble_ step to be where we get identified changpoints. They mentioned from "...Using the “assemble” keyword IsoSCM will assemble the mapped reads in a BAM file into a splice graph, identify nested terminal exons boundaries using the constrained segmentation procedure, and report the resulting models in GTF format....Pairwise comparison of tandem isoform usage can be performed using the “compare” keyword, which reports the relative usage of change points in each sample in a tabular format." 

Hence, it sounds like the _compare_ step outputs the identified change points (or PAS) that we want to report.

faricazjj:
Fixes #429 

**Background**
Currently, we get results for identification challenge from one of the output files from the _assemble_ step. Initially, this file seems to contain the identified changepoints just by looking at the content of the file.
Screen Shot 2022-07-26 at 11 32 38 AM

But as I read further beyond the assemble step, the second step they describe is the _compare_ command. The _compare_ command "reports the differential usage of each identified change-point", which I expected to show the site usages of the same sites as the sites from the previous _assemble_ step shown in the screenshot above. But I see less and also different sites in the _compare_ step output file
Screen Shot 2022-07-26 at 11 35 09 AM

This led me to look more into whether we should be extracting sites for identification challenge from the _assemble_ or _compare_ step.

As additional context, the steps to run isoscm are:
1. run _assemble_ step which creates a tmp folder: isoscm/tmp/{sample}.cp.filtered.gtf
2. run _compare_ step, this requires the xml file output from assemble step for two samples, but since we're getting site usage per sample, I put the same sample twice as input to obtain the following output:
Screen Shot 2022-07-26 at 12 18 54 PM

**Discussion**
Even though the first isoscm/tmp/{sample}.cp.filtered.gtf file contains changepoint locations, I'm not entirely sure we should obtain identification output sites from there since it's in a tmp folder and the github readme doesn't explain what the files in the tmp folders are--they only explained the files outside of the tmp folder from _assemble_ step i.e they explained the files from step 2 above but not files from step 1. I think the sites for identification might have to be obtained from the _compare_ step output (i.e. isoscm/compare/{sample}.txt). After reading their readme, I checked their paper and saw that in their paper, they didn't mention _assemble_ step to be where we get identified changpoints. They mentioned from "...Using the “assemble” keyword IsoSCM will assemble the mapped reads in a BAM file into a splice graph, identify nested terminal exons boundaries using the constrained segmentation procedure, and report the resulting models in GTF format....Pairwise comparison of tandem isoform usage can be performed using the “compare” keyword, which reports the relative usage of change points in each sample in a tabular format." 

**Outcome**
Hence, it sounds like the _compare_ step outputs the identified change points (or PAS) that we want to report.

## Checklist

- [ ] I have tested my code

### DIFF
--- a/execution_workflows/IsoSCM/modules/isoscm_assemble.nf
+++ b/execution_workflows/IsoSCM/modules/isoscm_assemble.nf
@@ -4,6 +4,13 @@ include { initOptions; saveFiles; getSoftwareName } from './functions'
 params.options = [:]
 def options    = initOptions(params.options)
 
+/*
+     IsoSCM assmeble step is the first step in IsoSCM. It constructs a splice-graph exons,
+     and segment terminal exons using the constrained change-point detection procedure.
+     One of the output files contain locations of change point that can be used for
+     APAeval identification challenge. Another output file is an XML file that is used for
+     the second step, the compare step.
+*/
 process ISOSCM_ASSEMBLE {
         publishDir "${params.outdir}/isoscm/isoscm_assemble_out_dir/${sample}", mode: params.publish_dir_mode
 	container "docker.io/apaeval/isoscm:latest"
@@ -14,15 +21,19 @@ process ISOSCM_ASSEMBLE {
 
         output:
         tuple val(sample), path(assemble_out_identification_rename), emit: ch_isoscm_assemble_out
+        tuple val(sample), path(assemble_out_xml_rename), emit: ch_isoscm_assemble_out_xml
 
         script:
-        assemble_out_identification_rename = "identification_out"
-        
+        assemble_out_identification_rename = "identification_out"       
       	assemble_out_identification = sample + ".cp.gtf"
-        
+        assemble_out_xml = sample + ".assembly_parameters.xml"
+        assemble_out_xml_rename = "assemble_out_xml"
+          
         """
         java -Xmx100G -jar /IsoSCM.jar assemble -bam $aligned_bam -base $sample -s $strand -dir isoscm
-
+        
         mv isoscm/tmp/$assemble_out_identification $assemble_out_identification_rename
+        mv isoscm/$assemble_out_xml $assemble_out_xml_rename
         """
+       
 }
--- a/execution_workflows/IsoSCM/modules/isoscm_compare.nf
+++ b/execution_workflows/IsoSCM/modules/isoscm_compare.nf
@@ -1,0 +1,33 @@
+// Import generic module functions
+include { initOptions; saveFiles; getSoftwareName } from './functions'
+
+params.options = [:]
+def options    = initOptions(params.options)
+
+/* 
+    ISOSCM compare is the second step of the tool. The splice-graphs outputs from the assemble step is used in the compare
+    step to perform joint segmentation of the terminal exons of coexpressed transcripts. This step reports the realtive usage
+    of identified change points in each sample in a tabular format
+*/
+process ISOSCM_COMPARE {
+        publishDir "${params.outdir}/isoscm/isoscm_assemble_out_dir/${sample}", mode: params.publish_dir_mode
+	container "docker.io/apaeval/isoscm:latest"
+        label 'process_high'
+
+        input:
+        tuple val(sample), path(assemble_out_xml_rename)
+
+        output:
+        tuple val(sample), path(relative_usage_quantification_out), emit: ch_isoscm_compare_out
+
+        script:
+        compare_out_relative_usage_quantification_rename = "relative_usage_quantification_out"
+        
+      	compare_out = sample + ".txt"
+        
+        """
+        java -Xmx100G -jar /IsoSCM.jar compare -base $sample -x1 $assemble_out_xml_rename -x2 $assemble_out_xml_rename
+        
+        mv compare/$compare_out $compare_out_relative_usage_quantification_rename
+        """
+}
--- a/execution_workflows/IsoSCM/modules/postprocess_identification.nf
+++ b/execution_workflows/IsoSCM/modules/postprocess_identification.nf
@@ -13,7 +13,7 @@ process POSTPROCESS_IDENTIFICATION {
         label 'process_high'
 
         input:
-        tuple val(sample), path(assemble_out_identification)
+        tuple val(sample), path(compare_output)
 
         output:
         path(output_file), emit: ch_postprocess_identification_out
@@ -23,6 +23,6 @@ process POSTPROCESS_IDENTIFICATION {
         output_file = sample + "_" + identification_out_suffix
         
         """
-        postprocess_identification.py $assemble_out_identification $output_file
+        postprocess_identification.py $compare_output $output_file
         """
  }
--- a/execution_workflows/IsoSCM/subworkflows/run_isoscm.nf
+++ b/execution_workflows/IsoSCM/subworkflows/run_isoscm.nf
@@ -5,6 +5,7 @@
 def modules = params.modules.clone()
 
 include { ISOSCM_ASSEMBLE } from '../modules/isoscm_assemble' addParams( options: [:] )
+include { ISOSCM_COMPARE } from '../modules/isoscm_compare' addParams( options: [:] )
 include { POSTPROCESS_IDENTIFICATION } from '../modules/postprocess_identification' addParams( options: [:] )
 
 workflow RUN_ISOSCM {
@@ -16,6 +17,11 @@ workflow RUN_ISOSCM {
         Run IsoSCM assemble step for each sample
      */
      ISOSCM_ASSEMBLE( ch_aligned_bam_files_dir )
+
+     /*
+        Run IsoSCM compare step for each sample
+     */
+     ISOSCM_COMPARE( ISOSCM_ASSEMBLE.out.ch_isoscm_assemble_out_xml )
 
      /*
         Get identification output file


mrgazzara:
@faricazjj I've looked into this a bit for the test data and the other mouse samples you ran locally. I agree with you that the output of the assemble step is not suitable. Spot checking through some of the coordinates, many seem to correspond to transcription start sites as well. This leads me to think it's some more general gene/transcript annotation derivative file and we cannot really tell what is what easily.

Using the compare step and parsing from the `relative_usage_quantification_out` is the way to go, BUT currently we are only grabbing half of the sites from this output (the proximal PAS only). The output of `relative_usage_quantification_out` looks like this: 
Screen Shot 2022-09-28 at 8 50 51 AM

We are grabbing, I think, the coordinate from `changepoint` column. This is great and corresponds to the changepoint within the broader region considered, meaning this is a proximal PAS only. The distal PAS can also be obtained from each row. The `locus` column represents the entire region examined by the tool (`upstream_segment` + `downstream_segment`) so to get the dPAS coordinate we just need to take the `end` coordinate from `locus` if `strand == +` and take the `start` coordinate from `locus` if `strand == -`. 

Maybe there's another issue on how to make this work for relative inclusion, but just continuing on from the above to get it written down, because each row represents a pPAS + dPAS combo we can grab the relative quantification information from here as well from the `site_usage` column (second to last). So `site_usage` should be the percent pPAS used (the `changepoint` coordinate above) and `1 - site_usage` would be the percent dPAS used (from the `locus` coordinate described above). 

I am not sure how / if the tool will report genes with single PAS only (no changepoint).

yuukiiwa (approved):
Thanks so much, @faricazjj!! Looks good to me!

faricazjj:
Thank you so much for the review and the fix @mrgazzara !! :D I added the extraction of distal PAS in the new commit so we're now getting two sites from each row, one for proximal and the other for distal. Here's an example for control_r1 of the test data:
Isoscm compare step raw txt output:
Screen Shot 2022-09-28 at 11 30 04 AM

Our workflow's identification output bed:
Screen Shot 2022-09-28 at 11 30 39 AM

Does this look right?

execution_workflows/IsoSCM/bin/postprocess.py
```diff
@@ -0,0 +1,70 @@
...
+        rd = csv.reader(fd, delimiter="\t", quotechar='"')
+        next(fd, None)
+        for row in rd:
+            # obtain proximal PAS
+            chrom = row[2].split(":")[0]
+            # need to do -1 to convert from 1-based gtf file coordinate to 0-based bed file coordinate
+            start = int(row[2].split(":")[1].split("-")[1]) - 1
+            end = start + 1
+            strand = row[8]
+            name = '|'.join([chrom, str(start) + ":" + str(end), strand])
```
mrgazzara, line 36: Small comment, but not worth blocking this for now as it is more important for relative usage quantification output:

For a tool like this we will need to keep track of which pairs of PAS go together when we match and correlate dPAS versus pPAS values. Because this tool only outputs coordinates (no gene or transcript info), it would be helpful to have the `name` here be unique to the pair of PAS that are derived from each line of this file. 

I was thinking the `locus` information (`row[7]`), but that is not unique in the mouse cortex dataset we ran so we might have to use that plus a line number or something. I'm thinking it should end up looking something like: 

[line_number]_[locus]_proximal and [line_number]_[locus]_distal

Again, not 100% necessary for this so feel free to ignore for now.

execution_workflows/IsoSCM/bin/postprocess.py
```diff
@@ -0,0 +1,70 @@
...
+            output = [chrom, str(start), str(end), name, ".", strand]
+            outputs.append(output)
+
+            # obtain distal PAS
+            # if positive strand, distal PAS is the end of the locus
+            if strand == '+':
+                start = int(row[7].split(":")[1].split("-")[1]) - 1
```
mrgazzara, line 43: Request changing `row[7]` to `row[6]`

Explanation: 
Doing some testing I have found that `locus` is not unique and is shared between some lines. These lines have more than 2 PAS / change points detected in the area. IsoSCM seems to tile pairs of PAS together in quantifying them and `locus` coordinates correspond to the most proximal to most distal region. This means we would grab the same `dPAS` from these lines using the coordinates from `locus` even though it does not correspond to the right distal region coverage in one of the two lines. Here's an example:
image

The solution is to just grab the dPAS coordinates from the `downstream_segment` column (index 6). They already handle strand so grabbing the coordinate here for dPAS will always be correct. Just grab the last coordinate if `+` and first coordinate if `-` (which you're already doing).

execution_workflows/IsoSCM/bin/postprocess.py
```diff
@@ -0,0 +1,70 @@
...
+            # if negative strand, distal PAS is the start of the locus
+            elif strand == '-':
+                start = int(row[7].split(":")[1].split("-")[0]) - 1
```
mrgazzara, line 46: Request changing `row[7]` to `row[6]` for the explanation given above for the comment on line 43.

mrgazzara (changes requested):
Looking good. Made a few comments in `postprocess.py` to fix the dPAS coordinates for certain edge cases I found with more testing.

execution_workflows/IsoSCM/bin/postprocess.py
faricazjj: Should this be row[5]? I see that when strand is -, the upstream segment's coordinates are bigger than downstream segment coordinates. Should we be taking the last coordinate of upstream segment (index 5) for - strand instead of the first coordinate of downstream segment (index 6)? Here is an example of a locus with 3 PAS from the examples I was printing:
Screen Shot 2022-09-29 at 9 36 00 PM

execution_workflows/IsoSCM/conf/modules.config
```diff
@@ -63,21 +63,21 @@ params {
                 If using APAeval test data, the file is provided in tests/test_data folder
                 e.g gtf_genome_file = '/home/fzhuang/apaeval/APAeval/execution_workflows/IsoSCM/../../tests/test_data/gencode_2genes_Chr_prefix.vM18.annotation.gtf'
             */
-            gtf_genome_file = '[path_to]/../../tests/test_data/gencode_2genes_Chr_prefix.vM18.annotation.gtf'
+            gtf_genome_file = '/home/fzhuang/new_apaeval/APAeval/execution_workflows/IsoSCM/../../tests/test_data/gencode_2genes_Chr_prefix.vM18.annotation.gtf'
```
dominikburri, line 66: really minor comment but you should keep the `[path_to]` in this file.

dominikburri (approved):
Hi @faricazjj, thank you for the fixes!
I ran the test files and checked the identification BED output in IGV and I think it looks good.
Only have minor comment, good to go from my side.

execution_workflows/IsoSCM/conf/modules.config
faricazjj: oh yes thank you! I accidentally pushed my local version as well, will revert this to the original version!

execution_workflows/IsoSCM/bin/postprocess.py
mrgazzara: @faricazjj no row index 6 is correct for both

dominikburri (approved):
LGTM, thank you!

mrgazzara (approved):
Looks good! Nice catch on the negative strand downstream segment coordinate stuff :)

execution_workflows/IsoSCM/bin/postprocess.py
faricazjj: Ah yes sure thing!! I'll make sure to incorporate this in the relative quant isoscm PR! :D

dominikburri (approved):
Thanks :)

PR #433, opened 2022-09-29
SWF: include parameter to filter expression

### LINKED ISSUE #424: AbsQuant SWF: filter for non-zero expression
Currently all sites reported by the participant that do not match a ground truth site are considered false positives. However, QAPA apparently reports back *all* sites it has encountered in its reference DB, but if it didn't find them expressed the score will be set to zero. By considering all non-matched PD sites false positives, for QAPA that number is thus overestimated.
This effect is not relevant for the "core" abs quant metrics, but as we also calculate precision and sensitivity it might still influence how we represent QAPAs performance.

Whether or not we should filter for non-zero sites should be open to debate though, as one could argue that this return of all reference DB sites is a design decision of the developers, and we'd give QAPA an advantage if we filtered out those with a particular (in this case zero) score.

Opinions welcome, the implementation shouldn't be complicated, small adjustment of `compute_metrics.py`. We'd however have to re-run the abs quant SWF for QAPA for all challenges.

dominikburri:
Fixes #424 

The filtering happens at two stages: 

- matched sites are filtered for prediction scores > `tpm_threshold`. This will be used for various metrics, including true positives and the various correlation metrics. 
- prediction sites are filtered for scores > `tpm_threshold`. This will be used for computing false positives.

Let me know if this makes sense in both cases. 

## Checklist

- [x] I have tested my code


faricazjj (approved):
thank you @dominikburri LGTM!! :D

mrgazzara (approved):
Looks great!

PR #437, opened 2022-10-06
[BugFix] QAPA error converting string "NA" to float when calculating PPAU

### LINKED ISSUE #436: Bug: QAPA error converting string "NA" to float when calculating PPAU
Thanks to @mrgazzara for reporting the bug!

**Problem**
An error occurred when running QAPA on Mayr data stating that string "NA" could not be converted to float in PPAU calculation.

Screen Shot 2022-10-06 at 7 33 22 PM

**Solution**
A solution to this is when the PPAU value is NA, we leave the string as is. Otherwise, we convert PPAU value to float when calculating PPAU fraction.

faricazjj:
Fixes #436 

Thanks to @mrgazzara for reporting the bug!

**Problem**
An error occurred when running QAPA on Mayr data stating that string "NA" could not be converted to float in PPAU calculation.

Screen Shot 2022-10-06 at 7 33 22 PM

**Solution**
A solution to this is when the PPAU value is NA, we leave the string as is. Otherwise, we convert PPAU value to float when calculating PPAU fraction.

The solution works 
Screen Shot 2022-10-06 at 7 41 40 PM

## Checklist

- [x] I have tested my code

### DIFF
--- a/execution_workflows/QAPA/bin/make_quant_bed.py
+++ b/execution_workflows/QAPA/bin/make_quant_bed.py
@@ -9,7 +9,12 @@ qapa_quant_bed_tpm=open(sys.argv[3],'w')
 
 for ln in qapa_results_file:
  ln=ln.strip().split('\t')
- ppau_fraction=float(ln[12])/100.0
+ # if the ppau value is NA, don't calculate ppau fraction
+ if ln[12] == "NA":
+  ppau_fraction = ln[12]
+ # if ppau is not NA, calculate ppau fraction by dividing by 100
+ else:
+  ppau_fraction=float(ln[12])/100.0
  if ln[7] == '+':
   chromStart,chromEnd=str(int(ln[9])-1),ln[9]
  elif ln[7] == '-':


dominikburri (approved):
Thank you @faricazjj, LGTM!

SamBryce-Smith (commented):
Nice spot @faricazjj, happy for this to be merged if comments not answered, I'm mainly asking out of curiosity.

Does the summary workflow validate that the score column of BED files can be cast to a float? Or more importantly, are rows with NA values ignored/dropped without causing any errors? If dropped/handled in the summary workflow I think it's fine to proceed as is. Otherwise I'm not sure I see a reason to report the rows in output if they have NA as we won't be able to do any useful metrics calculations with them (I'm also not sure how/why QAPA reports some entries with NA...).

Dropping rows with NA values (either here or in summary workflow) also shouldn't lead to 'orphan' PAS (e.g. only 1 site for a terminal exon/gene retained) due to how QAPA calculates PPAU (see [compute_pau.R](https://github.com/morrislab/qapa/blob/b7dd879edd063a114d9c5df9835d35c991390248/scripts/compute_pau.R#L82)). Because the default for base R `sum()` is to not remove NAs, if a single PAS for a terminal exon has an NA value all PAS for that TE will have it. For example:

```{r}
> x <- c(NA, 20, 30)
> x / sum(x) * 100
[1] NA NA NA
> x / sum(x,na.rm = T) * 100
[1] NA 40 60
```

faricazjj:
> Nice spot @faricazjj, happy for this to be merged if comments not answered, I'm mainly asking out of curiosity.
> 
> Does the summary workflow validate that the score column of BED files can be cast to a float? Or more importantly, are rows with NA values ignored/dropped without causing any errors? If dropped/handled in the summary workflow I think it's fine to proceed as is. Otherwise I'm not sure I see a reason to report the rows in output if they have NA as we won't be able to do any useful metrics calculations with them (I'm also not sure how/why QAPA reports some entries with NA...).
> 
> Dropping rows with NA values (either here or in summary workflow) also shouldn't lead to 'orphan' PAS (e.g. only 1 site for a terminal exon/gene retained) due to how QAPA calculates PPAU (see [compute_pau.R](https://github.com/morrislab/qapa/blob/b7dd879edd063a114d9c5df9835d35c991390248/scripts/compute_pau.R#L82)). Because the default for base R `sum()` is to not remove NAs, if a single PAS for a terminal exon has an NA value all PAS for that TE will have it. For example:
> 
> ```
> > x <- c(NA, 20, 30)
> > x / sum(x) * 100
> [1] NA NA NA
> > x / sum(x,na.rm = T) * 100
> [1] NA 40 60
> ```

@SamBryce-Smith thanks for pointing this out Sam and for showing the example!! I actually thought about this a little bit--whether we're filtering NA values here or in swf--but I think since we're filtering out 0's in swf, my impression is that we're going to filter out NAs there as well and so the ewf's output is just the raw values of what the tool gives. @ninsch3000 @dominikburri @yuukiiwa @mrgazzara any thoughts? are we handling NA values in the different swfs currently?

dominikburri:
Hm good point. I didn't check in detail, but the matching is done with bedtools: https://github.com/iRNA-COSI/APAeval/blob/e875dd20780252492c5350b017fc1f1b43059e20/utils/apaeval/src/apaeval/main.py#L97
And there the score doesn't matter.

For the quantification, at some point the prediction TPM is summed with python Series, here: https://github.com/iRNA-COSI/APAeval/blob/e875dd20780252492c5350b017fc1f1b43059e20/summary_workflows/quantification/quantification_dockers/q_metrics/compute_metrics.py#L130
this excludes NA, so you should always get non-NA values (except all is NA). 

For the correlation computation we use scipy.stats module, for which I don't know how they are dealing with this. But I assume NAs are simply skipped. 
I hope this helps a bit.

faricazjj:
@dominikburri I see, thank you for the clarification!! Sounds to me like it's fine to leave the NAs in ewf outputs, will merge this PR!

PR #441, opened 2022-10-14
Execution workflow update: IsoSCM doesn't run with reverse_forward and unstranded combination

### LINKED ISSUE #440: Execution workflow update: IsoSCM doesn't run with reverse_forward and unstranded combination
**Background**
IsoSCM allows for two settings in terms for strandedness--unstranded and reverse_forward. It wasn't explicitly stated in the tool's README but when we have single end reads with reverse/forward strandedness, we can't set IsoSCM's strandedness to be reverse_forward. This will cause an error. Hence, we need to add checks in the workflow to prevent users from using the wrong combination of inputs and having to wait for the tool to run before encountering this error.

**Solution**
Add a parameter check when reading the samplesheet to never have reverse_forward and single read combination

faricazjj:
Fixes #440 

**Background**
IsoSCM allows for two settings in terms for strandedness--unstranded and reverse_forward. It wasn't explicitly stated in the tool's README but when we have single end reads with reverse/forward strandedness, we can't set IsoSCM's strandedness to be reverse_forward. This will cause an error. Hence, we need to add checks in the workflow to prevent users from using the wrong combination of inputs and having to wait for the tool to run before encountering this error.

**Solution**
Add a parameter check when reading the samplesheet to never have reverse_forward and single read combination

## Checklist

- [x] I have tested my code

### DIFF
--- a/execution_workflows/IsoSCM/README.md
+++ b/execution_workflows/IsoSCM/README.md
@@ -34,8 +34,8 @@ columns:
 
 - sample: name of the sample (e.g control_replicate1)
 - bam: absolute path to BAM input file for the sample 
-- strand: the strandedness of the data, can be 'reverse_forward' or 'unstranded'
-- read_type: whether the sample contains paired-end or single-end reads, can be 'paired' or 'single'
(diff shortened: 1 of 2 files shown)


yuukiiwa (approved):
Thanks so much, @faricazjj!! Looks good to me!

dominikburri (approved):
Hi @faricazjj, thank you, LGTM!

PR #444, opened 2022-10-25
SWF: allow multiple challenges per infile

ninsch3000:
This fix is necessary if we want to compare an ewf ouput file to multiple ground truths, e.g. genome wide ground truth AND a filtered version with only terminal exons.    
The SWF selects the challenge name from a given list, depending on the ewf output file name. With this fix we allow multiple challenges to be selected if the first part of their name matches the challenge part of the ewf outfile name. Subsequently we loop over the selected challenges and validate the files or calculate the metrics, respectively.    

## NOTES
- in `compute_metrics.py` it is intentional that the looping is done *inside* the `compute_metrics` function, as `main.nf` expects a single output file from this step. Metrics for all challenges matching a given infile are therefore stored in the same assessment.json, which carries the name of the infile. If looping were done outside the function, i.e. the function were to be called several times, the assessment.json file would be overwritten and only results from one challenge stored.
- in `validation.py`, validation of input files is probably done multiple times, if multiple challenges are found... For the sake of time constraints this behaviour has not been checked or corrected.

### ewf output
`[toolname].[challengename].[eventcode].bed`   
example: `PAQR.MmusCortex_adult_R1.02.bed`

### ground truth
`[challengename].[challenge_keyword].[genome_id].bed`    
example: `MmusCortex_adult_R1.mm10.bed` and `MmusCortex_adult_R1.TE.mm10.bed`

### challenges
corresponding ground truth name without `.bed` extension

## Checklist

- [x] I have tested my code


ninsch3000:
## TO DO
* [ ] adapt identification SWF
* [ ] double check output jsons
* [ ] remove DEBUG print statements

dominikburri (commented):
Hi @ninsch3000, I tested the changes in `compute_metrics.py` manually and it seems fine. 
Thank you for adding an additional challenge.
One question though. Does the gathering of challenges in the line below require exact matching of list entries as provided in `tool_event.config`?
```
challenges = [c for c in challenge_ids if c.split('.')[0] in str(participant_input)]
```
I assume it wouldn't, as the right challenge is only selected by the challenge name, as you write in the description. Just want to be clear about that.

So if you can add the changes also for identification it would be good to go from my side.

mrgazzara:
Great idea. This should also give us flexibility to do more filtering on the ground truths, if necessary, in the future. 

Just wanted to raise one point about the example ground truth given:
> **ground truth**
> `[challengename].[challenge_keyword].[genome_id].bed`
> example: `MmusCortex_adult_R1.mm10.bed` and `MmusCortex_adult_R1.TE.mm10.bed`

Here the first example file (`MmusCortex_adult_R1.mm10.bed`) would be the unfiltered PAS bed with all ground truth sites (no filtering), but this one lacks a `[challenge_keyword]`. I assume we should also have one for this. Something like `all`? Open to suggestions.

ninsch3000:
Hey @mrgazzara , yes of course you can add `all`. The example was just to show that the `challenge_keyword` is optional. In the meanwhile I have created a more extensive specification in the main SWF README, please check it out. With the info there you should be all set to do the renaming of the GT, EWF out, and whatever other files you were going to remain :)

ninsch3000:
# NOTE
If you want to test the identification workflow, you will have to change `i_metrics/requirements.txt`, such that it uses branch `multiple_challenges_p_infile` to install the `JSON_templates` module!

dominikburri (approved):
Hi @ninsch3000, many thanks!
I ran locally the quantification test without problems and everything seems to be in place.
The identification bit also seems fine, the metrics are on the appropriate places, all 3 challenges are executed and the rogue metrics (thanks for dealing with this!) are also properly saved. 
So good to go from my side :)

mrgazzara (approved):
Everything looks good to me!

PR #446, opened 2022-11-02
Fix: utils metrics2tsv to work for identification event

ninsch3000:
When splitting metrics ids and expanding columns the script failed for identification jsons because some metrics don't contain window sizes, and there are no site sets in identification (script was written for AbsQuant files).   
Quick fix here: Replace `None` with mock value. Column `site_set` will contain mostly "NA", but "all" for metric AUC. This is necessary because pandas won't expand cols if there are only "NA" values.

## Checklist

- [ ] I have tested my code


dominikburri (approved):
LGTM! 
Just to be clear, as we discussed: this work when AUC is included in the metrics, then in the column _site_set_ a cell with content "all" will be included, the remaining cells will be "NA".

PR #451, opened 2022-11-08
Execution workflow: GETUTR - report Format 04 BED file in workflow (per-PAS fractional relative usage)

### LINKED ISSUE #450: Execution workflow: GETUTR - report Format 04 BED file in workflow (per-PAS fractional relative usage)
**Background**
GETUTR outputs 2 files - smoothed 3' UTR and poly(A) cleavage sites with relative usage per site. For identification challenge, we extract the output file containing poly(A) cleavage sites. An example of the output file is below:
Screen Shot 2022-11-08 at 1 31 32 PM

**Solution**
For relative quantification challenge, we could take the column containing relative usage per site. We want a relative quantification challenge output file that looks like the following:
Screen Shot 2022-11-08 at 1 35 54 PM
The name column includes a numbered suffix for each site to indicate the number of sites in each transcript.

Big thanks to @mrgazzara for the POC of this issue!

faricazjj:
Fixes #450 

**Background**
GETUTR outputs 2 files - smoothed 3' UTR and poly(A) cleavage sites with relative usage per site. For identification challenge, we extract the output file containing poly(A) cleavage sites. An example of the output file is below:
Screen Shot 2022-11-08 at 1 31 32 PM

**Solution**
For relative quantification challenge, we could take the column containing relative usage per site. We want a relative quantification challenge output file that looks like the following:
Screen Shot 2022-11-08 at 1 35 54 PM
The name column includes a numbered suffix for each site to indicate the number of sites in each transcript.

Big thanks to @mrgazzara for the POC of this issue!

**Results**
After incorporating relative quantification workflow, the new execution workflow runs successfully:
Screen Shot 2022-11-08 at 1 41 10 PM
The output files now contain both identification and relative quantification challenge output files:
Screen Shot 2022-11-08 at 1 41 30 PM

## Checklist

- [x] I have tested my code

### DIFF
--- a/execution_workflows/GETUTR/bin/postprocess_relative_quantification.py
+++ b/execution_workflows/GETUTR/bin/postprocess_relative_quantification.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+
+import sys
+import argparse
+
+
+def parse_args(args=None):
+    description = "Reformat GETUTR output bed file into the output file of relative usage quantification challenges"
+    epilog = "Example usage: python convert_to_bed.py <FILE_IN> <RUN_IDENTIFICATION> <RELATIVE_USAGE_QUANTIFICATION_OUT>"
+
+    parser = argparse.ArgumentParser(description=description, epilog=epilog)
+    parser.add_argument("FILE_IN", help="Input deAPA output txt file.")
+    parser.add_argument("RELATIVE_USAGE_QUANTIFICATION_OUT", help="Name of output file for relative usage quantification challenge")
+    return parser.parse_args(args)
+
+
+def reformat_bed(file_in, relative_usage_quantification_out):
+    """
+    This function reformats GETUTR bed output file to file for relative quantification challenges
+    :param file_in: bed file to be reformatted
+    :param relative_usage_quantification_out: relative usage quantification output file name
+    :return: N/A
+    """
+
+    relative_usage_quantification_outputs = []
+    
+    outfile = open(relative_usage_quantification_out, "w")
+
+    with open(file_in) as f:
+        curr_transcript = ''
+        count = 0
+        for line in f:
+            chrom = line.split('\t')[0]
+            start = line.split('\t')[1]
+            end = line.split('\t')[2]
+            transcript = line.split('\t')[3]
+            if curr_transcript == transcript:
+                count += 1
+            else:
+                curr_transcript = transcript
+                count = 0
+            name = transcript + '_' + str(count)
+            score = line.split('\t')[4]
+            strand = line.split('\t')[5]
+            row = [chrom, start, end, name, score, strand]
+            outfile.write("\t".join(row))
+    outfile.close()
+
+def main(args=None):
+    args = parse_args(args)
+    reformat_bed(args.FILE_IN, args.RELATIVE_USAGE_QUANTIFICATION_OUT)
+
+
+if __name__ == '__main__':
+    sys.exit(main())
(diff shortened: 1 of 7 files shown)


mrgazzara (approved):
Looks good to me! Output of the results look good.

execution_workflows/GETUTR/modules/postprocessing_relative_quantification.nf
```diff
@@ -0,0 +1,26 @@
...
+    container "docker.io/apaeval/getutr:latest"
+
+    input:
+    tuple val(sample), path(getutr_output)
+
+    output:
+    path "*"
+
+    script:
+    relative_quantification_output = "${sample}_relative_quantification_output.bed"
```
dominikburri, line 22: Can you make the file suffix a parameter in the config? By default we will likely save it as sample_04.bed.

execution_workflows/GETUTR/bin/postprocess_relative_quantification.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python
+
+import sys
+import argparse
+
+
+def parse_args(args=None):
+    description = "Reformat GETUTR output bed file into the output file of relative usage quantification challenges"
+    epilog = "Example usage: python convert_to_bed.py <FILE_IN> <RUN_IDENTIFICATION> <RELATIVE_USAGE_QUANTIFICATION_OUT>"
```
dominikburri, line 9: the parser only has two arguments. I guess <RUN_IDENTIFICATION> is not needed, correct?

execution_workflows/GETUTR/bin/postprocess_relative_quantification.py
```diff
@@ -0,0 +1,55 @@
...
+
+    with open(file_in) as f:
+        curr_transcript = ''
+        count = 0
+        for line in f:
+            chrom = line.split('\t')[0]
+            start = line.split('\t')[1]
+            end = line.split('\t')[2]
+            transcript = line.split('\t')[3]
+            if curr_transcript == transcript:
```
dominikburri, line 37: I think this assumes that the input file is sorted by transcript. Is this the case? 
If the transcript occurs after a while, then the counting would start at 0 again and they would not be discernible by name.

dominikburri (commented):
Hi @faricazjj, thank you!
I only had time to go over the code a bit, I could not test it yet. Wanted to share some thoughts.

execution_workflows/GETUTR/bin/postprocess_relative_quantification.py
faricazjj: yes their output is already sorted by transcript so it's safe to do this but you're right I'll change this so that we're always grouping by transcript to be a hundred percent sure

execution_workflows/GETUTR/bin/postprocess_relative_quantification.py
faricazjj: you're right!

execution_workflows/GETUTR/modules/postprocessing_relative_quantification.nf
faricazjj: sure!

dominikburri (approved):
Thank you @faricazjj!
I ran the test with `tests/test_data/siControl_R1_2genes_Chr_prefix.bam` and `tests/test_data/siSrsf3_R1_2genes_Chr_prefix.bam` using the corresponding gtf (`tests/test_data/gencode_2genes_Chr_prefix.vM18.annotation.gtf`) and it ran successfully! 
The results for identification and relative quantification seem reasonable. I see transcripts with 1, 2, 3 or even 4 PAS, and each sums to 1.
So good to go from my side! 😃

mrgazzara (approved):
Updates noted! Looks good to me.

PR #452, opened 2022-11-10
fix: swf-ident: reduce memory and time consumption 

dominikburri:
When dealing with big files, `compute_metrics.py` consumes a lot of memory and also takes very long to compute.
The reason for this is that for metric 5 computation, two list comprehensions are used to iterate over the genes in the genome. The output of the first list contains indices where a PAS was found. As this list is only used for computing how many PAS in a given gene are present, this does not need to be stored. 
The fix is to simply combine the two list comprehensions and only store the count of PAS per gene. This for sure decreases memory consumption and also should help in computation time.

## Checklist

- [ ] I have tested my code


faricazjj (approved):
LGTM thank you!

mrgazzara (approved):
Commuting so could not actually test it (yet) but your changes look fine to me!

PR #454, opened 2022-11-24
fix: filter on score only after matched modifications

dominikburri:
@ninsch3000 and I noticed when plotting scatter plots of correlation that the filtering by TPM threshold `-tpm` does not work as intended. 
The problem is that the function `split_pd_by_dist` can generate new prediction sites with score 0. The filtering was done prior to that though, so the matched dataframe ended up with 0 TPM scores even though the filter should remove them.

The fix is to move the filtering by TPM values to after modifying the matched object.

The function `merge_pd_by_gt` is not affected, since it would only merge and sum the expression from multiple prediction sites.

## Checklist

- [x] I have tested my code


ninsch3000 (approved):

SamBryce-Smith (approved):
Looks good to me thanks @dominikburri! Looking forward to seeing the updated results :)

PR #460, opened 2023-03-17
fix: make SWF output jsons OEB compatible

### LINKED ISSUE #459: SWF: make output jsons OEB compatible
*This issue concerns the summary workflow json outputs only, NOT the workflows themselves.*    

Currently our SWFs are not compatible with the OEB VRE infrastructure and must thus be run somewhere else. However, the output json files are supposed to comply with the OEB schemas, so that APAeval results can be uploaded into the OEB database and be visualized on the OEB website.   

  
## OEB schema validator:
- Either clone the repo https://github.com/inab/OEB_level2_data_migration , or fetch both the schema from https://github.com/inab/OEB_level2_data_migration/blob/master/oeb_level2/schemas/minimal_bdm_oeb_level2.yaml and the example file from https://github.com/inab/OEB_level2_data_migration/blob/master/minimal_dataset_examples/results_example.json
- Install our extended validator in a python virtual environment. The validator is available at pip https://pypi.org/project/extended-json-schema-validator/ .
- Run next command line to check all it is working:

 ```
ext-json-validate --guess-schema oeb_level2/schemas/minimal_bdm_oeb_level2.yaml minimal_dataset_examples/results_example.json
# Or
ext-json-validate --guess-schema minimal_bdm_oeb_level2.yaml results_example.json
```
​
- If it works, then you can start trying your own JSONs.

## Identified problems to be fixed:
* [x] json objects for challenges with lists of participants are missing in `consolidated_results.json`. Should look like:
```
{
        "id":"GTEXsim_cerebellum_R10.all.hg38_v26",
        "participants": [
            "PAQR"
        ]
    }
```
The `consolidated_results.json` is created in `merge_data_model_files.py`. The information for above described object is already present in `Manifest.json`, which, however, is not parsed by our consolidation scripts. Not sure where to add this, or which file or directory paths to change in `main.nf` so that `Manifest.json` would be parsed automatically.

* [x] time zone in validated.json is in wrong format, has to be one of   
```
from datetime import datetime, timestamp

datetime.now(timezone.utc).replace(microsecond=0).isoformat()

​# If you want to offer the timestamp using machine's time zone as reference, is almost the same:

from datetime import datetime, timestamp

datetime.now(timezone.utc).replace(microsecond=0).astimezone().isoformat()

​# But if you want to use time zones referenced by name, then you have to use pytz library.

import pytz
from datetime import datetime, timestamp

namedtz = pytz.timezone('Europe/Madrid')
datetime.now(timezone.utc).replace(microsecond=0).astimezone(namedtz).isoformat()

​
# All these alternative are providing correct timestamps, so you can use the one you prefer.

# (Side note) As the docker instance running the python code most probably does not have set up the host time zone (unless the environment variable TZ is passed from the host on instance creation), second snippet of code will return timestamps referenced to the time zone which was setup in the docker image.
```

### PR DESCRIPTION
**Pull request by ninsch3000**
Fixes #459, #461, #462, #463
Currently our SWFs are not compatible with the OEB VRE infrastructure and must thus be run somewhere else. However, the output json files are supposed to comply with the OEB schemas, so that APAeval results can be uploaded into the OEB database and be visualized on the OEB website.   
In this PR the following changes have been made:
- specify correct time zone in `JSON_templates`, remove timestamp from manifest object
- for all `participant` objects, that is for all `validated_participant.challenge.json`, only save a single `participant` object that contains all challenge ids from all `validated_participant.challenge.json` in the `consolidated_results.json`.
- include objects from Manifest, containing participant and challenge ids, in `consolidated_results.json`
- for identification: save rogue metrics jsons in same directory as validated and assessment jsons
- for identification: fix wrong github `JSON_templates` requirement link
- fix errors raised by linter
- fixed by OEB: error caused by `barplot` in aggregation objects

## Checklist

- [x] I have tested my code

### DIFF
--- a/summary_workflows/identification/nextflow.config
+++ b/summary_workflows/identification/nextflow.config
@@ -12,17 +12,17 @@ profiles {
    docker {
       process {
           withName: validation{
-            container = "docker.io/apaeval/i_validation:1.0"
+            container = "docker.io/cjh4zavolab/i_validation:1.0"
           }
       }
       process {
           withName: compute_metrics{
-            container = "docker.io/apaeval/i_metrics:1.0"
+            container = "docker.io/cjh4zavolab/i_metrics:1.0"
           }
       }
       process {
           withName: benchmark_consolidation{
-            container = "docker.io/apaeval/i_consolidation:1.0"
+            container = "docker.io/cjh4zavolab/i_consolidation:1.0"
           }
       }
       
--- a/summary_workflows/identification/nextflow.config
+++ b/summary_workflows/identification/nextflow.config
@@ -100,8 +100,6 @@ params  {
   genome_dir = "${params.data_dir}/ref/genomes" // Must contain a .gtf file with same substring as in challenge_ids, e.g. test.genome.mm10_test.gtf
   offline = 1 // type int; 0 for False
 
-  //set DEFAULT_eventMark
-  event_date = '2023-03-23'
 }
 
 // By default output execution reports
--- a/summary_workflows/identification/nextflow.config
+++ b/summary_workflows/identification/nextflow.config
@@ -80,7 +80,7 @@ params  {
 
   //name or OEB permanent ID for the benchmarking community
   community_id = "APAeval"
-  // name of APAeval benchmarking event (01: Identification, 02: absQuant, 03: diff, 04: relQuant)
+  // name of APAeval benchmarking event (Identification_01 or absQuant_02); only used for directory name, not passed to main.nf
   event = "Identification_01"
 
   // Boolean operator: if set to CLOSED the whole workflow is executed; if OPEN, metrics are computed but aggregation/consolidation is not performed
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@
 ### Code ###
 .vscode/*
 !.vscode/tasks.json
-!.vscode/launch.json
+.vscode/launch.json
 *.code-workspace
 
 ### Emacs ###
@@ -333,3 +333,5 @@ $RECYCLE.BIN/
 *.lnk
 
 # End of https://www.toptal.com/developers/gitignore/api/python,vim,code,vscode,emacs,sublimetext,linux,macos,windows
+
+DATA/
--- a/summary_workflows/JSON_templates/src/JSON_templates/main.py
+++ b/summary_workflows/JSON_templates/src/JSON_templates/main.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timezone
 import os
 import json
 import jsonschema
@@ -31,7 +31,7 @@ import sys
 """
 def write_participant_dataset( ID, community, challenges, participant_name, validated):
 
-    if validated == True:
+    if validated:
         status = "ok"
     else:
         status = "corrupted"
@@ -43,7 +43,7 @@ def write_participant_dataset( ID, community, challenges, participant_name, vali
         "type": "participant",
         "datalink": {
             "attrs": ["archive"],
-            "validation_date": str(datetime.now().replace(microsecond=0).isoformat()),
+            "validation_date": str(datetime.now(timezone.utc).replace(microsecond=0).isoformat()),
             "status": status
         },
         "participant_id": participant_name,
--- a/summary_workflows/identification/identification_dockers/build.sh
+++ b/summary_workflows/identification/identification_dockers/build.sh
@@ -11,7 +11,7 @@ if [ $# -gt 0 ]; then
 	tag_id="$1"
 
 	for docker_name in i_validation i_metrics i_consolidation ; do
-		docker build -t apaeval/"$docker_name":"$tag_id" "$docker_name"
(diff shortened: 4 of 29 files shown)

### CODE REVIEW

**Review by dominikburri — DISMISSED** [review]
Thanks @ninsch3000! 
Before testing, I cleared the singularity cache and ensured it uses the "cjh4zavolab" images. 
I tested identification and quantification with profile slurm, and it runs successfully.
The five first changes you list seem to be addressed :) The linter I did not check and the OEB related error I cannot assess.
I used `nextflow run main.nf -profile slurm -c tool_event.config` to run it.

However, there seems to be a problem with `params.participant_id`, which occurs two times, in nextflow.config and in tools_config. The execution is correct, but the results are saved under the value from nextflow.config (by default "DEF_PARTICIPANT"). I assume this is because the parameters from nextflow.config are read out, including the participant_id and the subsequent output directory (`output`), which can not be written again after the tool_config is read in. 
I could "fix" it by including `includeConfig "tool_event.config"` before the params specifications in nextflow.config and running `nextflow run main.nf -profile slurm` instead. 
I assume this is not part of this PR, so I just wanted to note it down for now, and we can deal with it somewhere else. If I overlooked another solution or fix, I'm happy to hear it.

Put together: the changes look fine for me, no breaking of the test file execution.

**Comment by dominikburri** [comment]
For the actual runs I did not observe this behaviour. It could be, that the processes write the file to the correct location. My previous comment and solution was not needed.

I had to change a line in metrics json2tsv, since the json now can have entries without "type", causing a KeyError. I caught it in the list comprehension by adding another if statement beforehand. 
The other commits are tailored to our execution.

File: summary_workflows/identification/nextflow.config [hunk 1]
```diff
@@ -66,10 +70,10 @@ params  {
   participant_id = "DEF_PARTICIPANT"
 
   // general data directory
-  data_dir = "$baseDir/APAeval_I_test_data"
+  data_dir = "$launchDir/../DATA"
```
ninsch3000, line 73: I wouldn't want to merge this, as it breaks compatibility with the APAeval test file paths

summary_workflows/identification/nextflow.config
```diff
@@ -66,10 +70,10 @@ params  {
...
   
   // directory where the 'gold standards' are found
-  goldstandard_dir = "${params.data_dir}/ref"
+  goldstandard_dir = "${params.data_dir}/ground_truths"
```
ninsch3000, line 76: I wouldn't want to merge this, as it breaks compatibility with the APAeval test file paths

summary_workflows/identification/nextflow.config
```diff
@@ -92,12 +96,12 @@ params  {
   otherdir = "${params.output}/other"
 
   // other parameters
-  windows = "100 50 10" // type int, several values inside string separated by spaces. The first value is used for relative PAS usage calculation.
-  genome_dir = "${params.data_dir}/ref/genomes" // Must contain a .gtf file with same substring as in challenge_ids, e.g. test.genome.mm10_test.gtf
+  windows = "100 50 25 10" // type int, several values inside string separated by spaces. The first value is used for relative PAS usage calculation.
+  genome_dir = "${params.data_dir}/genomes" // Must contain a .gtf file with same substring as in challenge_ids, e.g. test.genome.mm10_test.gtf
```
ninsch3000, line 100: I wouldn't want to merge this, as it breaks compatibility with the APAeval test file paths

summary_workflows/identification/nextflow.config
```diff
@@ -92,12 +96,12 @@ params  {
...
   offline = 1 // type int; 0 for False
 
   //set DEFAULT_eventMark
-  event_date = '2022-10-28'
+  event_date = '2023-03-23'
```
ninsch3000, line 104: I think here we have to make sure that all our runs from all events contain the same date, as otherwise it could be that they will be visualized in OEB separately (not compared)

ninsch3000:
# NOTE
Before merging, change branches in `requirements.txt` files to `main`

dominikburri (dismissed):
Hi @ninsch3000,
Thank you for the updates! 
Removal of unused stuff (i.e. AUC) calculation for sure makes sense. 

I ran the identification and quantification tests without problems. I beforehand removed the existing singularity images, ensuring that the new ones are pulled. 

So, LGTM from my side.

summary_workflows/identification/nextflow.config
```diff
@@ -80,7 +80,7 @@ params  {
 
   //name or OEB permanent ID for the benchmarking community
   community_id = "APAeval"
-  // name of APAeval benchmarking event (01: Identification, 02: absQuant, 03: diff, 04: relQuant)
+  // name of APAeval benchmarking event (Identification_01 or absQuant_02); only used for directory name, not passed to main.nf
```
dominikburri, line 83: you could duplicate this comment for quantification

dominikburri (approved):
Thanks @ninsch3000,
The tests run successfully and give the new event identifiers (instead of event date). 
I only have a minor request regarding a comment, but totally not needed.

dominikburri (approved):
LGTM! 
Many thanks @ninsch3000!

